PR #234, opened 2015-01-28
enable HAVE_STPCPY for arm/arm64 and x86

rockyzhang:
This is a function can be found in all android bionic libc, so, enables this will prevent the build failure for armeabi-v7 ABI.


madeye:
stpcpy is only available in the aarch64 and x86_64 targets. Or more specifically, stpcpy is not available in android targets below 5.0.  

http://androidxref.com/5.0.0_r2/search?q=stpcpy&defs=&refs=&path=&hist=&project=bionic

http://androidxref.com/4.4.4_r1/search?q=stpcpy&defs=&refs=&path=&hist=&project=bionic

PR #284, opened 2015-04-29
Add gradle build and using libsuperuser from maven

vmlinz:
Since I see a lot of people who want to play with android-shadowsocks are frustrated by the sbt build system and I found it not playing well with Android Studio with scala plugin, I make this project available as a gradle project which helps importing to Android Studio.

I spend some time to translate the build.sbt to build.gradle with the help of an plugin. Now anyone can import this project to Android Studio without errors.

1. Translate the build.sbt to build.gradle using https://github.com/saturday06/gradle-android-scala-plugin
2. Delete https://github.com/shadowsocks/shadowsocks-android/tree/master/src/main/java/eu/chainfire/libsuperuser from src code and using 'eu.chainfire:libsuperuser:1.0.0.+'
3. There are still some things that I am not sure, for example, the proguard settings and the version code for shadowsocks android.

If you are willing to accept this pull request or you need more refinement, please leave messages here.

Br,
Nick


vmlinz:
I think if I am going to delete libsuperuser from source, I have to add a corresponding dependency line in build.sbt

madeye:
1. As we're using a modified libsuperuser, you'd not remove it from the project.
2. I suggest to submit this pull request again with only gradle support.

vmlinz:
@madeye Ok, let me clean it up and submit later. Thanks for the explanation.

PR #299, opened 2015-05-14
update

vvvvcp:
fix some compile error


madeye:
You should not change this. 1.3.22 just works well.

PR #311, opened 2015-06-06
scan json config

haobug:
in most case, we got a json config file from the ss service seller,
this patch enable one to scan the qr bar code generated from that kind of json config file
by qrencode or other tools 

this make scan function of ss android compliance more with end-user expectation.


madeye:
The providers should publish their service info in URI format, rather than JSON format. As a result, I prefer not to merging this change.

Anyway, thanks for your pull request.

haobug:
there might be some mis-explanation.
it's not replace the URI way, this patch is an addition to the URI way.

PR #346, opened 2015-08-18
add tcp_nagle_disable in tun2socks

clowwindy:
Turn on TCP_NODELAY for TCP sockets in lwIP stack.

Notice: I haven't tested on Android yet.


madeye:
I think it should work. Thanks!

PR #376, opened 2015-10-03
Migrate to Android 6.0 (1)

Mygod:
There's still one issue left: `net.simonvt.menudrawer`. I might redesign the UI for this...


madeye:
You do not need to redesign the UI, just add the source code of `android.utils.FloatMath` to `src/java`java.

Mygod:
Well that library is obsolete so we need to remove it, sooner or later. :-P

PR #410, opened 2015-11-06
Update a lot of dependencies

Mygod:
I think we should move dependencies in jni directories to separate repos so that we can pull updates, bugfixes, etc. more easily. I've moved 2 of them.


madeye:
I will give you the permission to shadowsocks repos, Then change the namespace of subprojects to github.com/shadowsocks and submit the pull request again.

Mygod:
Okay.

PR #412, opened 2015-11-07
Add more material design (1)

Mygod:
There's still more I'd like to do. But here's the first draft. Screenshot will be soon added...


Mygod:
http://imgur.com/v7uzKFA

madeye:
Great, looks quite good.

Mygod:
Todo list:

1. The icons are quite ambiguous. I believe I've seen connect/disconnect icons somewhere. I will add a long-press toast but I won't replace them since I'm lazy to create one.
2. Test on 4.x devices.
3. Circular progress around the fab on connect/disconnect. (this will use the 3rd icon I added in this commit)
4. Replace the obsolete MenuDrawer. (with DrawerLayout/another ListPreference?)
5. Update some of the preferences to match Material Design Guidelines.

I will be working on other things for today so you can do these if you like to. :stuck_out_tongue:

PR #413, opened 2015-11-08
Add more material design (2)

Mygod:
There are also some feature (non-UI) updates in this PR (the last commit to be specific) including:

* Support switching to NAT/VPN without restarting the app;
* Remove the useless `isRoot` item in the `SharedPreferences`;
* Respect 4.x rooted user's NAT/VPN preference. (https://github.com/shadowsocks/shadowsocks-android/commit/4906ce8af58a87ea4b3536aa9cf7a6bf68c55d56#diff-0caa3c8457106211bc36b29f3381a9b1L536)

There's one last thing left to do that is to get rid of `MenuDrawer`. :smile:


madeye:
LGTM.

madeye:
About the floating action button, I suggest to use two background colors to present the statea: grey for disconnected, green for connected / connecting / disconnecting,

Mygod:
Okay. But I'm working on `MenuDrawer` now so here's the [color](https://www.google.com/design/spec/style/color.html#color-color-palette).

    material_grey_600: #757575

PR #414, opened 2015-11-08
Add more material design (3)

Mygod:
No more major updates from now on (I think). Btw it kinda looks like Chainfire's style now. :stuck_out_tongue: 

http://imgur.com/Yib9Hjs


madeye:
I think we need a drawer to put reset/flush dns/about and other actions into it. The actions like reset/flush dns should not be visible unless the user want to diagnose any issues.

madeye:
Also, NAT mode is already deprecated so that it's designed to be hidden from normal users.

Mygod:
However, I use flush dns all the time. Otherwise I would get SSL certificate invalid error.

VPN mode is fine but it won't remove the "!" sign in my status bar so I prefer NAT mode.

madeye:
You got DNS pollution issue just because you're using NAT mode... It's a limitation of NAT mode. Also, without ROOT, NAT mode should not be visible.

To remove the `!` icon, please refer to this: https://www.noisyfox.cn/45.html

Mygod:
NAT mode is disabled instead of hidden on devices without root so that users will know why they can't enable that.

We can add a summary to keep normal users from using that feature. The default mode on 5.0+ devices is VPN mode.

Since resetting and flushing dns isn't really destructive, I prefer to keep them at the top for convenience's sake.

Mygod:
Or we can add a miscellaneous category at the bottom and put flush/reset/about/faq/support... there.

madeye:
Sounds a good idea. Let me change the order.

PR #420, opened 2015-11-10
Switch to another profile while connected

Mygod:


Mygod:
Note: There are some TODOs that I've left in the source code. I'm probably not fixing them though. You can fix them if you like.

https://github.com/shadowsocks/shadowsocks-android/blob/master/src/main/scala/com/github/shadowsocks/preferences/DropDownPreference.scala#L36
According to [Material Design Guidelines](https://www.google.com/design/spec/components/menus.html#menus-simple-menus), the currently selected menu item aligns vertically over the list item. But system settings doesn't implement that anyway.

https://github.com/shadowsocks/shadowsocks-android/blob/master/src/main/scala/com/github/shadowsocks/ProfileManagerActivity.scala#L144
Reordering profiles by dragging. Requires database update. (store an order field maybe?)

https://github.com/shadowsocks/shadowsocks-android/blob/master/src/main/scala/com/github/shadowsocks/ShadowsocksSettings.scala#L13
Move preference related logic to `ShadowsocksSettings`. Just for clarity's sake.

madeye:
I think you can wait for pull requests from others (like what I am doing)  :laughing:

Mygod:
Sounds hopeless.

PR #431, opened 2015-11-18
Fix #430

### LINKED ISSUE #430: UI Problems on Android 4.x
screen shot 2015-11-18 at 3 33 29 pm

@Mygod Is there a overlay on the main UI?

Mygod:
TODO: Test to see if it works.

### DIFF
--- a/src/main/scala/com/github/shadowsocks/ParserActivity.scala
+++ b/src/main/scala/com/github/shadowsocks/ParserActivity.scala
@@ -39,15 +39,15 @@
 
 package com.github.shadowsocks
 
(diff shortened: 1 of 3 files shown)


madeye:
It seems we need also to extend all the DialogPreference...

What about just using native style for all dialog on devices below 5.0?

Mygod:
Sounds good.

P.S. http://stackoverflow.com/questions/29608126/dialog-skinning-with-appcompat-v7-22-results-in-ugly-shadows-on-api-21#comment47363209_29608126

PR #433, opened 2015-11-18
Fix #432

### LINKED ISSUE #432: Copy feature settings from the current profile to the new profile
I think we can assume that users will apply same feature settings for most of their profiles. Having this feature will save some time.

Mygod:


Mygod:
Update: added the total traffic statistics into the profile selector. (backend not implemented though)

Issue #440, opened 2015-11-21
Show traffic stats in notification bar


Mygod:
Now's a good time to add back this feature.

P.S. It seems I've removed `service_status` in this commit because it wasn't used: https://github.com/shadowsocks/shadowsocks-android/commit/51e155c940b2c35bb1d68910af3bc65d19d63183#diff-787c9f20c9297db53ee0a28a2f1cfab5L57

madeye:
This feature introduced too many issues, like system UI crashing. Unless we find a better way to show traffic stat on notification, I prefer to not adding this.

Mygod:
I assume this is what's causing the crashes: https://github.com/shadowsocks/shadowsocks-android/blob/v2.2.2/src/main/res/drawable/ic_stat_speed.xml

We can just display the rate in the notification text, and add display rate in notification bar with [this new API in API level 23](http://developer.android.com/reference/android/app/Notification.Builder.html#setSmallIcon(android.graphics.drawable.Icon)).

madeye:
The new API seems promising, which enables us creating icon from bitmap dynamically.

BTW, there already exists a helper function `Utils.getBitmap` to generate bitmap of a string.

Mygod:
Actually I think the design in #447 looks quite fine. Closing this.

PR #447, opened 2015-11-24
Show traffic in notification

Mygod:
This is actually what I had in mind when opening #440.

TODO: Disabling this new feature causes weird behaviors:

1. The change doesn't apply immediately;
2. Shutting down service doesn't update UI normally.

I can't figure out why or how to fix it though. It seems like something about inter-process things. :-(

### DIFF
--- a/src/main/res/values-zh/strings.xml
+++ b/src/main/res/values-zh/strings.xml
@@ -32,8 +32,7 @@
   <string name="bypass_apps_summary">启用该选项，以使所选应用程序的流量不经过代理</string>
   <string name="auto_connect">自动连接</string>
   <string name="auto_connect_summary">随系统启动后台服务</string>
-  <string name="traffic_stat">流量统计</string>
-  <string name="traffic_stat_summary">在通知栏上显示当前上下行速率</string>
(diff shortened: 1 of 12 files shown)


madeye:
I think it's safe to keep traffic stat always enabled.

Mygod:
Traffic stat is always enabled. But I decided to let user pick whether they want to show the current tx/rx rate in notification or not. However that doesn't work well...

madeye:
I suggest not to add more settings. The UI is already too complicated... Let's keep post speed to the notification.

Mygod:
Alright. Removed it for now. Please review the code. I did a force close and a reset to make it work so there might be more bugs...

madeye:
OK, I will check them later.

madeye:
What are the steps to reproduce "Shutting down service doesn't update UI normally."? I tried the change locally and it works very well.

madeye:
OK, the issue is already fixed by removing the "disable the new feature" option...

Mygod:
Yes. Checkout the previous commit to see the bug.

PR #456, opened 2015-11-30
Automation app integration

CzBiX:
Support automation app(Tasker, Locale) to turn on/off service.
I will add more feature(such as import profile, switch profile) if this PR is acceptable.


madeye:
Thanks, merged.

Mygod:
Importing profile sounds useless. Why would you want to automate that?

phpbestlang:
Hi @Mygod, I believe it's useful, like in different place/network can switch to different profile, or disconnect after enter a no-firewall network and reconnect after leave.

CzBiX:
@Mygod You are right, I just thought of this, but not seriously thought about it.
For most of user, switch service state and switch profile should be enough right now.

PR #457, opened 2015-11-30
Refine layout

Mygod:


src/main/AndroidManifest.xml
```diff
@@ -80,9 +80,7 @@
                 android:parentActivityName=".Shadowsocks"
                 android:launchMode="singleTask"/>
 
-        <activity android:name=".TaskerActivity"
-            android:icon="@drawable/ic_launcher"
```
CzBiX: Tasker will warn about icon missing, so this line shouldn't be removed.

Mygod: @CzBiX Okthx.

PR #459, opened 2015-12-01
add switch profile feature for Tasker

CzBiX:


Mygod:
It works. Looks awful though. :mask:

PR #467, opened 2015-12-05
Cache proxied apps for even better performance

Mygod:
Also some other improvements.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,0 @@
-sbt.version=0.13.8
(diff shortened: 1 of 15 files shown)


Mygod:
Here comes the real fix. Workaround removed. :laughing:

Mygod:
P.S. FYI, the problem is caused by these two lines:
https://github.com/Mygod/shadowsocks-android/commit/8c2690c0fb1c4dbc3aecb98a9cc2f6df4e097c54#diff-10a9b03b7f7592d0af1b344f99abb2a4R76
https://github.com/Mygod/shadowsocks-android/commit/8c2690c0fb1c4dbc3aecb98a9cc2f6df4e097c54#diff-10a9b03b7f7592d0af1b344f99abb2a4R76 (originally `attachService(null)`)

madeye:
Great!

PR #481, opened 2015-12-15
Fix #479 (HTTP client test)

### LINKED ISSUE #479: HTTP client test
There's a feature called HTTP client test in Phone info (you can access it via entering `*#*#4636#*#*` in dialer) which makes a simple HTTP request (to `www.google.com` I suppose) to test the connection. Is it acceptable if I add an item in Shadowsocks under (or above?) the traffic stats to do the similar thing to see whether a profile is usable, i.e. has Internet access?

Mygod:
I also added a useful util: `CloseUtils`, just for my own convenience.

### DIFF
--- a/src/main/res/values-zh/strings.xml
+++ b/src/main/res/values-zh/strings.xml
@@ -78,6 +78,11 @@
   <plurals name="bytes">
     <item quantity="other">字节</item>
   </plurals>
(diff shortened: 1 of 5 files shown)


Mygod:
UPDATE: Doesn't seem to work with NAT mode though.

PR #503, opened 2015-12-29
Reduce battery usage - Add option to disable notification traffic

Mygod:
This commit reverts "Removed the switch for now", commit 1eda4c47a4abba26f4669967f2c00a1ffca1452e in #447.

Fixed all the IPCs, now it's working fine.

According to my profiling results, `Notification.Builder.build` and `NotificationManager.notify` causes ~2x extra battery usage EACH even after I've done so many optimizations IIRC so it's reasonable to give users an option to turn this additional feature off.

Speaking of IPCs, is it possible to move everything in `ss-local`, `ss-tunnel`, etc. to process `:vpn` and `:nat`? Creating so many additional processes seems unnecessary (and battery draining).

### DIFF
--- a/src/main/aidl/com/github/shadowsocks/aidl/IShadowsocksService.aidl
+++ b/src/main/aidl/com/github/shadowsocks/aidl/IShadowsocksService.aidl
@@ -12,4 +12,6 @@ interface IShadowsocksService {
 
   oneway void start(in Config config);
   oneway void stop();
+
+  oneway void showNotificationTraffic(boolean value);
 }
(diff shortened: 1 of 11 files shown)


madeye:
Nope, we need separate processes to keep things simple, which prevents any potential crashes and memory leak. Also, I believe they have nothing to do with battery usage.

Actually, I didn't notice any visible battery drain with latest versions. If you looking into the CPU time consumed by Shadowsocks, it's very small.

madeye:
To verify this, you can also try Shadowsocks on a Android 4.4 or below device, it should report almost zero battery usage of shadowsocks.

Mygod:
Alright. Closing it for now.

madeye:
Let's collect more data and see what will happen.

CzBiX:
貌似升级到 2.9.x 之后，耗电量显著上升。
现在长期在电池榜排行第一，用量超过第二名屏幕两倍。之前都是第4，5名开外的。
不确定是否有关，供参考。

Kongloli:
之前的版本没在意，今天发现2.9.5耗电排行榜也是第一。

PR #544, opened 2016-01-19
Support DNS override using custom hosts file 

lucifer9:
For users without root


madeye:
Sorry, we do not have any plan to support this feature.

BTW, your code has two issues:
1. You need to upgrade database after adding a new column to the database.
2. Hosts file will be ignored when UDP relay mode is enabled.

yebr:
希望可以重新考虑，如果可自定义hosts会方便很多。

PR #549, opened 2016-01-21
sleep in TaskerReceiver to avoid race condition

CzBiX:


madeye:
Thanks. I implemented this using handler.postDelayed().

PR #574, opened 2016-02-16
Replace Future.onFailure with Future.onComplete

JLLK:
`Future.onFailure` has been deprecated since scala 2.12, let's use `Future.onComplete` instead.

In `scala.concurrent.Future`:
```
  @deprecated("use `onComplete` or `failed.foreach` instead (keep in mind that they take total rather than partial functions)", "2.12")
  def onFailure[U](@deprecatedName('callback) pf: PartialFunction[Throwable, U])(implicit executor: ExecutionContext): Unit = onComplete {
    case Failure(t) =>
      pf.applyOrElse[Throwable, Any](t, Predef.conforms[Throwable]) // Exploiting the cached function to avoid MatchError
    case _ =>
  }
```

### DIFF
--- a/src/main/scala/com/github/shadowsocks/package.scala
+++ b/src/main/scala/com/github/shadowsocks/package.scala
@@ -2,12 +2,16 @@ package com.github
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+import scala.util.{Failure, Try}
 
 /**
   * @author Mygod
   */
 package object shadowsocks {
-  def ThrowableFuture[T](f: => T) = Future(f) onFailure {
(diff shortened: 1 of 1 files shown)


Mygod:
scala 2.12 requires JDK 1.8. However Android currently uses JDK 1.6 and 1.7 from 4.4 on. So there is no plan to migrate to 2.12 for now. But seeing the implementation is no different from 2.11, I think it's okay to merge this.

> `keep in mind that they take total rather than partial functions`

And I think your code will throw `MatchError` if Future completes normally.

JLLK:
@Mygod  thank you , you're right 
I push a new commit to avoid the MatchError.

madeye:
Merged, thanks!

PR #583, opened 2016-02-25
Fix binder ipc issue when process died unexpected

JLLK:
If **shadowsocks:vpn** died unexpected, such as:

```bash
> adb shell ps | grep shadowsocks
u0_a52    1456  141   537172 40864 ffffffff b754e7ca S com.github.shadowsocks
u0_a52    1477  141   523808 32712 ffffffff b754ff37 S com.github.shadowsocks:vpn
u0_a52    1523  1477  6980   1424  c01d7f5f b7df5f37 S /data/data/com.github.shadowsocks/ss-local
u0_a52    1524  1477  8508   1236  ffffffff b7df577f S /data/data/com.github.shadowsocks/pdnsd
u0_a52    1525  1477  6960   1420  c01d7f5f b7df5f37 S /data/data/com.github.shadowsocks/ss-tunnel
u0_a52    1533  1477  6488   1180  c01d7f5f b7df5f37 S /data/data/com.github.shadowsocks/tun2socks
```

```bash
> adb shell kill 1477
```

We'll find **shadowsocks** won't respond to `trafficUpdated` or `stateChanged` no matter how many times we click on `FloatingActionButton`, because `bgService` in **shadowsocks** don't work since binder **died**. 

Though **shadowsocks:vpn** will restart immediately, **shadowsocks:vpn** can't send msg through binder ipc to **shadowsocks** without `bgService.registerCallback(callback)`, so I did something to fix:

1. registered `ShadowsocksDeathRecipient` to `bgService`.
2. let **shadowsocks** re-bind to `ShadowsocksVpnService`(or `ShadowsocksNatService`) when `binderDied` called.

And it works fine now.  :yum:

### DIFF
--- a/src/main/scala/com/github/shadowsocks/ServiceBoundContext.scala
+++ b/src/main/scala/com/github/shadowsocks/ServiceBoundContext.scala
@@ -4,16 +4,15 @@ import android.content.{ComponentName, Context, Intent, ServiceConnection}
 import android.os.{RemoteException, IBinder}
 import com.github.shadowsocks.aidl.{IShadowsocksServiceCallback, IShadowsocksService}
 import com.github.shadowsocks.utils.Action
-import com.github.shadowsocks.utils.Utils
 
 /**
   * @author Mygod
   */
-trait ServiceBoundContext extends Context {
-
+trait ServiceBoundContext extends Context { outer =>
   class ShadowsocksServiceConnection extends ServiceConnection {
     override def onServiceConnected(name: ComponentName, service: IBinder) {
       bgService = IShadowsocksService.Stub.asInterface(service)
+      bgService.asBinder().linkToDeath(new ShadowsocksDeathRecipient(outer), 0)
       registerCallback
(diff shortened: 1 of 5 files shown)


madeye:
LGTM.

@Mygod Any thoughts?

madeye:
Tested and verified locally.

Merged, thanks!

Mygod:
Well... We can merge this but a few issues needs fixing.

`IOUtils` seems redundant and flawed (including the block that invokes its method). Please note that we do have a [`CloseUtils`](https://github.com/shadowsocks/shadowsocks-android/blob/master/src/main/scala/com/github/shadowsocks/utils/CloseUtils.scala) in this project. Also the method name `inSafe` implies that it's completely safe to invoke this. But I think we shouldn't ignore exceptions whenever we want even if we logged them.

There are also a few issues in `ProcessUtils`. `inShadowsocks` is buggy. `com.github.shadowsocks:vpn/nat` starts with `com.github.shadowsocks`.

There are more minor issues. I think I will fix them when I've got time.

PR #585, opened 2016-02-26
Fix a issue in ProcessUtils

JLLK:
Fix a issue in ProcessUtils

### DIFF
--- a/src/main/scala/com/github/shadowsocks/ShadowsocksDeathRecipient.scala
+++ b/src/main/scala/com/github/shadowsocks/ShadowsocksDeathRecipient.scala
@@ -17,7 +17,7 @@ class ShadowsocksDeathRecipient(val mContext: ServiceBoundContext)
 
     mContext match {
       case ss: Shadowsocks =>
-        inShadowsocks {
+        inShadowsocks(mContext) {
           ss.unregisterCallback
           ss.bindToService()
         }
--- a/src/main/scala/com/github/shadowsocks/utils/ProcessUtils.scala
+++ b/src/main/scala/com/github/shadowsocks/utils/ProcessUtils.scala
@@ -2,6 +2,7 @@ package com.github.shadowsocks.utils
 
 import java.io.{BufferedReader, FileInputStream, InputStreamReader}
 
+import android.content.Context
 import com.github.shadowsocks.utils.IOUtils._
 
 /**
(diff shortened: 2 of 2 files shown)


Mygod:
Acceptable but clumsy.

madeye:
In which situation, we need this `pName.startsWith(SHADOWNSOCKS) && pName.length > SHADOWNSOCKS.length && pName.charAt(SHADOWNSOCKS.length) != ':')`?

Mygod:
https://github.com/shadowsocks/shadowsocks-android/pull/583#issuecomment-188777766

madeye:
Should `pname == SHADOWSOCKS` be enough?

madeye:
Also, why not get the process name though Android API? For example: http://stackoverflow.com/questions/8542326/android-how-to-get-the-processname-or-packagename-by-using-pid

Mygod:
I prefer this solution.
http://stackoverflow.com/a/21389402/2245107

madeye:
@Mygod OK, so there is a bug of ActivityManager.
@JLLK Could you refine your solution following the link provided by @Mygod?

Mygod:
@madeye I prefer this one just because getting all processes is an overkill imo.

And keep using `/proc/self/cmdline`.

JLLK:
There is a trick like this:

```
> adb shell cat /proc/1822/cmdline
/data/data/com.github.shadowsocks/tun2socks--netif-ipaddr26.26.26.2--netif-netmask255.255.255.0--socks-server-addr127.0.0.1:1080--tunfd48--tunmtu1500--sock-path/data/data/com.github.shadowsocks/sock_path--loglevel3--dnsgw26.26.26.1:8153%
```

So we have to use `startsWith` to judge process rather than the whole name.

Mygod:
It's actually NULL (`\0`) terminated/separated so just follow that link please.

madeye:
Forget to mention: Please avoid any`/data/data/com.github.shadowsocks` in your code. Use `getApplicationInfo.dataDir` instead.

IMO, if you just want to identify the JVM processes like `com.github.shadowsocks`or `com.github.shadowsocks:vpn`, a simple comparison should be enough.

madeye:
Merged. Thanks!

Mygod:
Much better. :smile:

PR #594, opened 2016-03-07
Recover unexpected process exit on MIUI

### LINKED ISSUE #499: MiUI7开发版12.10后台内存清理自动结束tun2socks、ss-local
在MiUI上后台管理界面上滑锁住应用防止自动清理后，MiUI自带清理依然会结束tun2socks及ss-local导致无法继续使用。明前尚不明确是针对SS进行的定点清除还是MiUI逻辑问题（因为只有tun2socks及ss-local被杀掉）

ayanamist:
fix #499
I'm not familiar with scala, so use Java instead to implement some classes.
Tested on Xiaomi 4c Dev 6.3.3


ayanamist:
It has a timeout, it will sleep 1 second and then check process which prevents fast-fail.
I learned scala a little before, and don't like its present situation. Why insisting on writing Scala?

madeye:
Yes, I noticed the timeout. A better way is to stop restarting if it's a fast-fail, to avoid battery drain.

Sorry, but this project only accepts pull requests in Scala because I'm a fanboy of Scala. :smile:

ayanamist:
OK, i changed to stop loop when fast-fail.
I also changed to scala. The converted code can not compile, so i waste hours to make it compiling, and ran into [this bug](https://issues.scala-lang.org/browse/SI-8162) so i have to extract a method `initThread`
Now i hate scala ever than before, and will prevent its usage in all my projects.

madeye:
Merged, thanks! I will refine the code later.

CzBiX:
@ayanamist You can take a try with Kotlin as a better replacement.

ayanamist:
@CzBiX I know Kotlin as a more simpler than Scala and better than Java programming language, however, this project is only receiving scala pr as above mentioned.

Mygod:
@madeye Well I suppose we could've merged this and done the java-to-scala conversion ourselves.

madeye:
@Mygod Yes, we can do this for further pull requests.

PR #598, opened 2016-03-11
Add quick switch from notification.

AnyOfYou:

PR #602, opened 2016-03-12
Automatic service restart (with zero latency) after switching profile

Mygod:
`start(Config)` and `stop()` has been replaced by `use(Config)` and `use(null)` to implement this feature. Invoking `start(Config)` with a different config will restart the service.

## A possible issue/race condition

There are three ways to switch profiles which is when service restart is required:

1. via Profile selector.
2. via Tasker.
3. via Quick Switch introduced in #598.

When the latter two are performed and `Shadowsocks` is resumed, it will sometimes start the service twice and after a chain of events it results in `Unable to connect to remote server` so the service stops itself.


Mygod:
And this should fix #537.

madeye:
Avoid `stopSelf()` when switching profiles via 017dd42.

PR #609, opened 2016-03-15
Fix parser bug, add share profile by NFC.

AnyOfYou:
The old setMessage(profiles.mkString("\n")) made profiles iterator point to end before it being use.

Use Android Beam or NFC tag to add profile.


Mygod:
Using toList or sth like that would be better imo.

AnyOfYou:
The problem is also about showing user string representation of Profile object is not useful.

Mygod:
Oops I meant to write `profiles.map(Parser.generate).mkString("\n")`.

madeye:
Merged, thanks!

PR #615, opened 2016-03-19
Add strings zh-TW

LNDDYL:

PR #624, opened 2016-03-29
Refine #614, #615

Mygod:


src/main/scala/com/github/shadowsocks/ProfileManagerActivity.scala
```diff
@@ -111,12 +102,9 @@ class ProfileManagerActivity extends AppCompatActivity with OnMenuItemClickListe
       }
     }
 
-    def onClick(v: View) = {
+    def onClick(v: View) {
       ShadowsocksApplication.switchProfile(item.id)
       finish
-      if (isLauncherFromShareIntent){
-        TaskStackBuilder.create(ProfileManagerActivity.this).addNextIntentWithParentStack(NavUtils.getParentActivityIntent(ProfileManagerActivity.this)).startActivities()
```
AnyOfYou: When user launch ProfileManagerAcitivity from share intent, after he/she added profile and clicked to the new profile, just simplely finish the activity may exit the app, user probably want to back to Shadowsocks activity to enable it. I think we should keep this logic or use something else instead (like automatic enable it when launch from share intent).

Mygod: If you want that, we can simply reverse #614. ProfileManagerActivity is a sub-activity of Shadowsocks. Don't break the hierarchy. If user want to switch profile, we now have quick switch.

PR #630, opened 2016-04-03
New UI polishment

### LINKED ISSUE #508: Proposal for the new UI polishment
The new material design style UI is great.
Might be it could be polished a little like below:
ss

* The switch button will be much easier to reach.
* Info structure is cleaner
* While scroll down, the Nework Trafiic part will be hidden.

Mygod:
Closes #508. (finally!)

screenshot_20160403-190806

The animation in this PR is a placeholder (again). I wrote an `ExpandCollapseAnimation` but for some reason `getMeasuredHeight` returns an incorrect value of 720 on my device. You can check it out [here](https://github.com/Mygod/shadowsocks-android/tree/height-animation) and try to figure out what's wrong. It seems like a bug of Android or support library but I'm not sure. :confused:


AnyOfYou:
We should tweak the layout of 'Network Traffic'
screen shot 2016-04-04 at 02 27 19

Mygod:
@AnyOfYou You're right. Will do in the next PR.

madeye:
What about showing the profile name on the title?

Mygod:
Not necessary. You can view/change profile name in Profile name. Showing it again would be redundant.

madeye:
Maybe we need to provide some hints here. People could get confused again with this UI change.

Mygod:
Umm a down arrow was added at the top. It's common in app design imo.

PR #628, opened 2016-04-01
Refine UI.

AnyOfYou:

PR #631, opened 2016-04-03
FAB ripple and usability problem fixes for #628

AnyOfYou:


madeye:
Maybe need to resolve some conflicts first.

Mygod:
Thanks. I'll fix the remaining problem.

PR #644, opened 2016-04-14
Support one-time authentication in ss urls for sharing

Mygod:
This commit implements the new ss protocol proposed here: https://github.com/shadowsocks/shadowsocks-org/issues/16


madeye:
Great! Merged.

AnyOfYou:
Geart, but is this also affect other shadowsocks client?

Mygod:
Not yet. I'm considering implementing this in shadowsocks-windows client later today.

PR #646, opened 2016-04-16
Add support for Android TV

Mygod:
This commit adds an app banner in Android TV and introduces several changes including refactoring AppManager to make all features accessible with D-pad navigation.

Currently this app has the exact same user interface on Android TV as on normal Android devices but one should create a TV-specific user interface in the future.

P.S. Only tested in Android emulator. More testing might be needed.

Screenshots attached.

screenshot_20160416-110908
screenshot_20160416-111533
screenshot_20160416-111615


madeye:
Looks awesome. Thanks!

PR #666, opened 2016-04-30
Use Chinese (Taiwan) translation as fallback for other Chinese locales

Mygod:
#615 makes Shadowsocks display in English for Chinese locales if it is neither Chinese (China) nor Chinese (Taiwan).

I choosed Chinese (Taiwan) for other Chinese locales. Here's my reasons.

* In my device (CyanogenMod 13), the only other Chinese locale I have is Chinese (Hong Kong) which uses traditional Chinese characters.
* According to [Wikipedia](https://en.wikipedia.org/wiki/Simplified_Chinese_characters#Education), traditional Chinese characters are more commonly used by Chinese native speakers outside of mainland China.

What do you think? @madeye @LNDDYL


LNDDYL:
我的意見是：「影梭」這個名稱只用於中文(中國大陸)[zh-CN]上，其他語系均使用其英文名稱「Shadowsocks」。

LNDDYL:
@Madeye @Mygod 為什麼把zh-TW移除了？

madeye:
It makes zh-TW the default locales of Chinese, not removing it.

LNDDYL:
意義何在？

PR #676, opened 2016-05-06
add a perl script to generate chnroute

wongsyrone:
The perl version takes two seconds while the python version takes more than 30 seconds

Usage: perl ./gen.pl < chnroute.txt > chnroute-merged.txt


wongsyrone:
The time is measured when generate chnroute in the previous PR, the number of raw data line is 7808.

PR #725, opened 2016-06-19
Improve translations for export toast

kaneawk:
Most of guys does not know export to where.


Mygod:
Incorrect translation for traditional Chinese.

kaneawk:
:)

PR #742, opened 2016-07-04
Remove .aidl.Config and stop persisting SharedPreferences

Mygod:
Previously a profile can exist in the following forms:

* `.database.Profile` used for everything backend;
* `.ShadowsocksSettings` which uses Preferences to provide UI to tweak settings;
* `.aidl.Config` used for IPC;
* `<data dir>/shared_pref/com.github.shadowsocks_preferences.xml` which exists for no apparent reason.

Thus long code block can be seen whose purpose is simply converting the data from one form to another. This commit intends to remove the latter two by:

1. Using profile ID and take advantage of fs r/w lock SQLite uses for IPC and better extensibility;
2. Stop persisting redundant fields in the preferences file and update database directly.

Important note: everything should work but a thorough testing is recommended.


madeye:
LGTM.

PR #765, opened 2016-07-15
enable ipv6 in libudns

wongsyrone:


madeye:
No DNS queries happens in `shadowsocks-android`. All of them are handled in `pdnsd` instead.

PR #788, opened 2016-08-05
Add KCP support

madeye:
Add KCP support for shadowsocks.

KCPTUN should be setup on a different port to avoid conflicts with Shadowsocks' UDP support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,8 @@ out/
 #NDK
 src/main/obj
 src/main/libs
-src/main/assets/armeabi
-src/main/assets/armeabi-v7a
-src/main/assets/x86
-src/main/assets/api-16/
+src/main/assets/armeabi-v7a/*
+!src/main/assets/armeabi-v7a/kcptun
+src/main/assets/x86/*
+!src/main/assets/x86/kcptun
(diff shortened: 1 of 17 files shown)


codexss:
希望在服务端也集成这个功能

cj1324:
👍

Mygod:
Shouldn't we provide a more reasonable default value for [`kcpcli`](https://github.com/shadowsocks/shadowsocks-android/pull/788/files#diff-e538750d345d58bb7a056d5e3b86d09dR109), like `--crypt none --sndwnd 1024 --mode fast --nocomp`?

Mygod:
@madeye What do you think?

madeye:
I'm doing experiments for best parameters. Currently, I'm using `--crypt none --mode normal --mtu 1200 --nocomp --dscp 46 -datashard 10 -parityshard 0`

Mygod:
@madeye Should we use `--crypt none --mode normal --mtu 1200 --nocomp --dscp 46 -datashard 10 -parityshard 0` as the default value for now?

madeye:
@Mygod Yes, I think so.

PR #793, opened 2016-08-06
Fix remove problem

kaneawk:
Should use the `rm -rf` command to remove a directory.


wongsyrone:
This is a wrong fix, in my case, only kcptun and tun2socks will be copied to asserts/ dir.
You can compare
https://travis-ci.org/shadowsocks/shadowsocks-android/builds/150279490#L6810
and
https://travis-ci.org/shadowsocks/shadowsocks-android/builds/150275444#L6814.

The final apk size decreased from 6.67MB to 5.56MB. It should not happen.

kaneawk:
I can see that. madeye fixed this.

PR #799, opened 2016-08-09
Update dependency

wongsyrone:


Mygod:
Build failed: https://travis-ci.org/shadowsocks/shadowsocks-android/builds/150808901#L1158

madeye:
Let's revert this commit for now.

wongsyrone:
OK, lack of testing.

wongsyrone:
Maybe the correct cmdline is:
```
 script:
     - $SBTPATH/sbt android-update all
     - $SBTPATH/sbt native-build android:package-debug
```
I'm not sure atm.

Mygod:
Personally I find this useless since `native-build` doesn't depend on Android SDK.

Also, the current script does update the necessary components to the latest version.

wongsyrone:
Just want to suppress sbt warning.

wongsyrone:
If sbt supports updating SDK and other google/android tools, then we can drop self-written scripts.

Mygod:
[It is not self-written script.](https://github.com/shadowsocks/shadowsocks-android/blob/master/.travis.yml#L27)

Also, you can see [right here](https://travis-ci.org/shadowsocks/shadowsocks-android/builds/150522431#L6790) that `sbt-android` is able to download Android support repository dependencies anyways. Updating all SDK and all that time it's going to take would be a pain in the ass.

madeye:
I think we may still follow Mygod's scripts.

wongsyrone:
Well, all depends on @madeye.
Just wondering whether the versions in travis docker image will get out-dated. If travis stands at the cutting edge, each way is OK.

Mygod:
Yes. The original SDK version is so outdated that we need to update `tool` first so that it can discover the up-to-date xml file which contains Build Tools 24+ which is required for vector drawable-related stuff.

And `extra-google-m2repository` is required since [`sbt` can't discover them by itself](https://travis-ci.org/shadowsocks/shadowsocks-android/builds/150522431#L6814).

madeye:
@wongsyrone Please open another pull request with only updating the plugin. Remember to revert the update-sdk commit.

Mygod:
@madeye @wongsyrone No need, I will do it.

wongsyrone:
OK, I will test my opinion on my fork, if I have good news, I will report back.

PR #809, opened 2016-08-16
Add support for TCP Fast Open

Mygod:
TODO: `/etc/init.d` isn't supported by most custom ROMs but not all the devices. For example, Android emulator doesn't support `/etc/init.d`. Maybe some better approach to solve this?

### DIFF
--- a/src/main/scala/com/github/shadowsocks/utils/TcpFastOpen.scala
+++ b/src/main/scala/com/github/shadowsocks/utils/TcpFastOpen.scala
@@ -1,0 +1,39 @@
+package com.github.shadowsocks.utils
+
+import java.io.File
+
+import eu.chainfire.libsuperuser.Shell
+
+/**
+  * @author Mygod
+  */
+object TcpFastOpen {
+  /**
+    * Is kernel version >= 3.7.1.
+    */
+  lazy val supported = "^(\\d+)\\.(\\d+)\\.(\\d+)".r.findFirstMatchIn(System.getProperty("os.version")) match {
+    case Some(m) =>
+      val kernel = m.group(1).toInt
+      if (kernel < 3) false else if (kernel > 3) true else {
+        val major = m.group(2).toInt
+        if (major < 7) false else if (major > 7) true else m.group(3).toInt >= 1
+      }
+    case _ => false
+  }
+
+  def sendEnabled = {
+    val file = new File("/proc/sys/net/ipv4/tcp_fastopen")
+    file.canRead && (Utils.readAllLines(file).toInt & 1) > 0
+  }
+
+  def enabled(value: Boolean) = if (supported) {
+    val fastopen = if (value) 3 else 0
+    Shell.SU.run(Array(
+      "mount -o remount,rw /system",
+      "echo '#!/system/bin/sh\nsysctl -w net.ipv4.tcp_fastopen=" + fastopen + "' > /etc/init.d/tcp_fastopen",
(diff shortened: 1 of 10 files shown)


madeye:
Cool!

src/main/scala/com/github/shadowsocks/utils/TcpFastOpen.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def enabled(value: Boolean) = if (supported) {
+    val fastopen = if (value) 3 else 0
+    Shell.SU.run(Array(
+      "mount -o remount,rw /system",
+      "echo '#!/system/bin/sh\nsysctl -w net.ipv4.tcp_fastopen=" + fastopen + "' > /etc/init.d/tcp_fastopen",
+      "chmod 755 /etc/init.d/tcp_fastopen",
+      "sysctl -w net.ipv4.tcp_fastopen=" + fastopen,
+      "mount -o remount,ro /system",
+      "echo done"))
```
madeye: This line looks really dangerous. Maybe we should warn the user before enabling TCP fast open ?

Mygod:
P.S. I tested it on emulator by enabling tfo with and root, and also tested boot script creation on my device by running a custom build (because unfortunately my device is still using kernel 3.4.110).

src/main/scala/com/github/shadowsocks/utils/TcpFastOpen.scala
Mygod: Which line to be specific?

madeye: The line add a config to `/etc/init.d`. If anything goes wrong, it may cause soft-brick.

Mygod: Oh I see. I think I can do sth about that.

Mygod:
How about this? If this isn't enough, we can still add an additional warning message.

madeye:
One quick question: why we need that init script? Should modifying `/proc/sys/net/ipv4/tcp_fastopen` be enough to enable tfo?

Mygod:
If we don't have that init script, the change will be reverted when user reboots its device.

madeye:
But we always modify `/proc/sys/net/ipv4/tcp_fastopen` before establishing a VPN connection, do we?

Mygod:
Yep that's also an option. But that will cause Shadowsocks ask for root permission more often. I prefer using an init script because that makes it feel like turning on a system settings. :stuck_out_tongue:

madeye:
OK. Let's do this for now. If we found anyone soft-bricks his phone, we may try other approach. :laughing:

Mygod:
Yep. For further references, if somebody soft-bricks his phone, reboot into recovery, do an `adb shell rm -f /etc/init.d/tcp_fastopen` to un-brick your phone and report!

madeye:
Also, what about `echo 1 > /proc/sys/net/ipv4/tcp_fastopen` instead of using `sysctl`. There are always some weird phones without this kind of tools.

Mygod:
Ok. Are you sure we should use 1 instead of 3 which also enables receiving tfo?

madeye:
It's a typo, Yes, we need `3` here.

Mygod:
```
root@generic_x86:/ # sysctl
/system/bin/sh: sysctl: not found
```
Just tested on Android 4.4 emulator. Switching to echo is a bright idea.

PR #823, opened 2016-08-20
KCP with session ID

madeye:
Add session ID support with `--sid` option in the command line.

### DIFF
--- a/kcptun/make.bash
+++ b/kcptun/make.bash
@@ -51,7 +51,7 @@ export PATH=$GOROOT/bin:$PATH
 pushd kcptun/client
 
 echo "Get dependences for kcptun"
-go get -u github.com/xtaci/kcp-go
+go get -u github.com/madeye/kcp-go
 go get
 
 echo "Cross compile kcptun for arm"


Mygod:
Okay I've got it working now. Sorry for my procrastination. :stuck_out_tongue: Let's see if it works stably.

By the way, I found out that some routers block UDP packets. Do you plan to do some kind of fallback for that?

Mygod:
Also, how about adding `--sid` to default CLI parameter? This will only affect users upgrading from 3.0.3 (or 4, I can't quite remember).

madeye:
For routers blocking UDP, the user have to keep using shadowsocks over tcp.

The session ID is still not stable. To keep the compatibility with the upstream of kcptun, we'd better not make it as default option.

In the future, if this feature is merged (or refined) by upstream, we can make it enabled by default.

Mygod:
I'm still experiencing connection loss with sid enabled. 😕

madeye:
Could you try this commit? https://github.com/madeye/kcp-go/commit/ee66db5f9513401f7a20c3df8d5d0713d8b1d9c2

Simply rebuild kcptun in shadowsocks will apply this change.

Mygod:
Okay testing...

Btw wouldn't this kind of dependency management break [reproducible builds](https://reproducible-builds.org)?

Mygod:
Ok... So the connection can work without network listener but with a delay. Also connection loss problem is still present.

madeye:
It seems the connection loss problem is not related to the "connection loss" but a "connection frreezing" issue.

By running tcpdump on both client and server, I noticed several things:

1. Every UDP packet is sent and received sucessfully. It's verfied in tcpdump
2. Frequently strems are open and closed on the server. Open a TCP connection on the client leading to a stream open on the server,
3. Redialing works well when the "connection freezing" happens. The source addr:port changes on the server correctly.
4. No yamux i/o timeout. It seems that yamux thinks the underlying connection works correctly.

I'm thinking it may be a bug of KCP or yamux, causing freezing connection.

madeye:
The pain point is that I cannot reproduce this issue stablly. It just happens from time to time.

Mygod:
From my experience, it stops working every time after a while (~5-15 min after initial connection).

madeye:
One interesting finding: if I disable KeepAlive in Yamux, no more connection loss can be observed until now (2 hours).

https://github.com/shadowsocks/kcptun/blob/shadowsocks/client/main.go#L359

Also notice that @xtaci is trying to improve KeepAlive in Yamux. Not sure if they are related.

https://github.com/xtaci/yamux/commit/76ceb0030b43cf52acd9b8b1a36fa7ef5514a285

xtaci:
it's a bug in yamux。I'm trying to fix that.

madeye:
What about just disabling Keep-alive in yamux? Since we already have similar feature for NAT keep-alive, is keep-alive in yamux is still useful? @xtaci

xtaci:
absolutely ， diabling keepalive will cause memory leak for long running server. and i'm trying to find a way to do ping with keepalive and without performance penalty.

xtaci:
besides i prefer application layer keepalive which will be the most accurate. transport layer keepalive is not the best solution. so yamux is prefered

madeye:
OK, looking for your fixes.

xtaci:
https://github.com/hashicorp/yamux/pull/39  @madeye  
I've made a PR to yamux , and the bug description lies here.

Mygod:
Just then I tried https://github.com/shadowsocks/kcptun/commit/0e926339561163da15bfb87698b81cd8205a130c. And connection was lost after ~39 min. From the log I can see that at that time the server is still able to receive packets but it seems no packets are able to reach the client.

madeye:
With session ID or not? And after 1min, could this connection be restored?

Mygod:
With sid. Maybe I should try it without sid?

No.

madeye:
It still looks like an issue caused by either yamux or kcp, since the underlying UDP connection works well.

madeye:
Could you try to disable SID and KeepAlive of yamux? It may cause many non-closed sessions on server, but we can at least tell if it's still an issue of KeepAlive.

Mygod:
Still the same (connection lost).

Mygod:
P.S. I have disabled keep alive via setting `EnableKeepAlive:        false` both on client and server.

madeye:
I think we finally get a workaround by setting the `autoexpire` option (e.g. 60 sec).

Since we still suspect the freezing issue comes from yamux, by making kcptun not that "mux", we could avoid reusing the freezing yamux session. https://github.com/xtaci/kcptun/issues/159

Please update to the latest commit and try again. BTW, I dropped the SID support, which looks too tricky.

Also, I'm thinking do we really need yamux? As shadowsocks works as kind of "short connection" proxy, each KCP session could just be one TCP connection of shadowsocks. If so, no more freezing and even no more QoS.

madeye:
Or we may just use a more aggressive `autoexpire` number, says 10 sec, then we should find one trade-off between mux and non-mux.

PR #866, opened 2016-09-07
support Android 6 with PIE flag enabled

dna2github:
Android 6 support binaries only with PIE flag enabled.
Thus add the flags.

Tested on Huawei Honor 5A with Android 6: works.

Thanks,
Seven Liu


madeye:
You don't need to add that flag manually. All the binaries are built against android-16.

See https://github.com/shadowsocks/shadowsocks-android/blob/master/src/main/jni/Application.mk#L2 for more details.

PR #1004, opened 2016-12-13
acl: add telegram IPs

wongsyrone:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:


wongsyrone:
I just found http://bgp.he.net/search?search%5Bsearch%5D=Telegram&commit=Search. Should others been added?

madeye:
Not sure. It seems telegram is not using all of its IPs for chatting service.

wongsyrone:
I see.

PR #1011, opened 2016-12-15
Use symlink for gfwlist.acl

### LINKED ISSUE #1008: Duplicate acl files?
I just noticed that there's an [acl dir](https://github.com/shadowsocks/shadowsocks-libev/tree/master/acl) in shadowsocks-libev repo as well. Is it necessary to keep two copies of acls?

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

gfwlist.acl is moved to shadowsocks-libev and a symlink is created here in this repo. Fixes #1008.

Build verified.


madeye:
Great! So, it's a symbolic link to shadowsocks-libev repo.

Mygod:
Wait I just realized this still breaks backwards compatibility. See [here](https://raw.githubusercontent.com/shadowsocks/shadowsocks-android/master/src/main/assets/acl/gfwlist.acl).

madeye:
Hmm, let's revert it for now....

Mygod:
I think we can merge this just before the next stable release and suggest everyone using GFW list AND [v3.2.5-v3.3.1](https://github.com/shadowsocks/shadowsocks-android/commit/0191d1d0fc1e707c9ef6613a9edf308e73aa71ff) to upgrade asap. What do you think?

madeye:
IMO, it's too risky. Let me update both repos and keep them consistent in the future.

Mygod:
Okay. Meanwhile let's migrate to a better solution while we wait for these versions to die out. For example, using urls like `https://shadowsocks.org/acl/android/v1/gfwlist.acl` which is just 302 redirection to `raw.githubusercontent.com`?

wongsyrone:
What a great idea.

madeye:
@Mygod LGTM. Just add https://shadowsocks.org/acl/android/v1/gfwlist.acl to cloudflare.

Mygod:
Yes. Don't forget the other files in acl dir.

PR #1062, opened 2017-01-07
Let Android system extract native libraries

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

Advantage:

* No more worries about inconsistencies about native libraries and extracting them (therefore no more ETXTBSY errors);
* Native libraries are counted in App section instead of Data section in storage information.

Disadvantage:

* Executable names have to be `lib*.so` otherwise Android system won't extract it.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -7,19 +7,9 @@ function try () {
 # Build native binaries
 pushd src/main
 try $ANDROID_NDK_HOME/ndk-build -j8
-
+# Clean up old binaries (no longer used)
 rm -rf assets/armeabi-v7a
 rm -rf assets/x86
-mkdir -p assets/armeabi-v7a
-mkdir -p assets/x86
-
-#copy executables
-for app in pdnsd redsocks ss-local ss-tunnel tun2socks
-do
-    echo $app
-    try mv libs/armeabi-v7a/$app assets/armeabi-v7a/
-    try mv libs/x86/$app assets/x86/
-done
 popd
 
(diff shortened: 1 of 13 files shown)


madeye:
It looks quite interesting. I didn't know we can do this before.

wongsyrone:
I think

> Executable names have to be lib*.so otherwise Android system won't extract it.

should be included into commit message

Mygod:
Another advantage is that we don't need to chmod by ourselves. However because of its disadvantage we should be cautious about merging this. What do you think?

madeye:
I don't think the disadvantage matters. Is this approach also works on Android 4.4?

Mygod:
Only tested on Android 7.0.

madeye:
OK, let me test it first.

madeye:
Tested on Android 4.4. Everything works fine. Let's merge this change.

Mygod:
I noticed that after this change, installing debug apk takes 5 seconds to 1 minute longer. I wonder if it's caused by some kind of Android optimization or gdb_server.

madeye:
@Mygod The release apk seems to work fine. No delay observed.

wongsyrone:
Might related to https://developer.android.com/studio/command-line/zipalign.html

Mygod:
zipalign is not relevant since it's done in the compilation process.

PR #1070, opened 2017-01-09
Speed up ndk-build from sbt-android

wongsyrone:

### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,4 +50,4 @@ install:
       fi
 
 script:
-    - $SBTPATH/sbt native-build android:package-debug
+    - $SBTPATH/sbt -DNPROC=$(nproc) native-build android:package-debug
--- a/build.sbt
+++ b/build.sbt
@@ -8,6 +8,11 @@ name := "shadowsocks"
 version := "3.4.1"
 versionCode := Some(173)
 
+// Speed up ndk-build by reading nproc from environment
+// and default to 8
+val nproc = sys.props.getOrElse("NPROC", default = "8")
+ndkArgs := Seq(s"-j$nproc")
+
 platformTarget := "android-25"
 
 compileOrder := CompileOrder.JavaThenScala


madeye:
It's a very clever change. 😀

wongsyrone:
You can also modify README to clarify its usage

Mygod:
Last 7 successful builds took 644 ± 11s. This commit took 634 ± 2s (it built 2 times). I'm not convinced this actually speed up the build given ndk build takes quite some time.

wongsyrone:
Yeah. You can simply revert it.

Mygod:
Based on my observations, all my 8 processors are ~95% busy when doing the native build without specifying nproc manually. Why did you think this will speed up the build?

madeye:
It's quite interesting. I didn't expect ndk could do parallel building without `-j` options. If so, this change may be not necessary.

madeye:
Let's revert it for now. Anyway, thanks for this patch.

wongsyrone:
You win. Bye.

Mygod:
I seemed to be wrong. I observed again today and it seems 100% CPU usage seems to be caused by JNI build and JVM build running in parallel. And when JVM build finished none of my cores are 100% busy so there is probably room for improvement. Could you please redo this fork on our develop branch (which removed kcptun and builds 25% faster) and see if you can do some fine tuning and obtain some noticeable speed improvements (either on Travis or user's machine)? Thanks!

Mygod:
Okay I just took a look and the proper way to do this turned out to be this: `ndkArgs := "-j" :: java.lang.Runtime.getRuntime.availableProcessors.toString :: Nil`. There isn't a `$NPROC` on my OS, only `$(nproc)`.

# Benchmark

Test command: `android:package` (after `clean`)

* Before adding: CPU usage is relatively low after Proguard finishes. Elapsed time: 86s.
* After adding: CPU usage is ~100% even after Proguard finishes. Elapsed time: 46s.

So that's a 40s gain! :+1:

I'm pushing this change to develop branch and let's see how well it performs on Travis...

wongsyrone:
LGTM. The initial idea is just keep consistent with the previous build commands.

PR #1087, opened 2017-01-20
Move config files to files directory

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

It looks nicer.

Other changes btw:

* Removes duplicated logcats from GuardedProcess;
* local_port is moved to command line options since that's not sensitive;
* pdnsd uses cache dir;
* Refinements to Utils.

`shadowsocks-libev` is updated. Merge `no-prefix` branch if this gets merged.

### DIFF
--- a/mobile/src/main/scala/com/github/shadowsocks/BaseService.scala
+++ b/mobile/src/main/scala/com/github/shadowsocks/BaseService.scala
@@ -20,7 +20,7 @@
 
 package com.github.shadowsocks
 
-import java.io.IOException
+import java.io.{File, IOException}
 import java.net.InetAddress
 import java.util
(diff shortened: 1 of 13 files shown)


Mygod:
Note: We can also remove prefix options from the plugins. Plugins can simply use `protect_path` in their cwd.

madeye:
Hmm.. How to merge the change to submodule?

Mygod:
Merge this branch: https://github.com/shadowsocks/shadowsocks-libev/tree/no-prefix

Oh I just realized I forgot to change getopt for ss-tunnel. Do that for me when merging please. :stuck_out_tongue:

PR #1102, opened 2017-01-28
Add an option on local address

DKingCN:
### Type of changes

* [ ] Bugfix
* [x] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

Hi, this is a little tweak `adding an option on local_address`. So the ss-local/ss-tunnel will be able to expose its service to somewhere, mainly used in lan.

As you can see the commit is simple enough so no more explanations.

This can be useful because many linux distributions lacks built-package of ss-libev ,especially when python ss could be extremley slow and unstable with massive connections and without acl, that will be even hard to pull the needed source to build ss-libev. Like the problem of chicken or egg.
That is the situation I am up to these days. So I figured this out to solve it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ local.sbt
 .idea/
 out/
 .d
+*~
 
 #NDK
 src/main/obj
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,6 @@ local.sbt
 .idea/
 out/
 .d
-*~
 
 #NDK
 src/main/obj
--- a/src/main/scala/com/github/shadowsocks/ShadowsocksVpnService.scala
+++ b/src/main/scala/com/github/shadowsocks/ShadowsocksVpnService.scala
@@ -142,7 +142,7 @@ class ShadowsocksVpnService extends VpnService with BaseService {
       case None => throw NameNotResolvedException()
     }
 
-    handleConnection()
+    handleConnection()///////////////////////////////////////;;;;;;;;;;;;;;;;;;;;;;;;;
     changeState(State.CONNECTED)
 
     if (profile.route != Acl.ALL && profile.route != Acl.CUSTOM_RULES)
@@ -198,14 +198,14 @@ class ShadowsocksVpnService extends VpnService with BaseService {
 
   def startShadowsocksUDPDaemon() {
     val conf = ConfigUtils
-      .SHADOWSOCKS.formatLocal(Locale.ENGLISH, profile.host, profile.remotePort, profile.localPort,
+      .SHADOWSOCKS.formatLocal(Locale.ENGLISH, profile.host, profile.remotePort, profile.localAddress, profile.localPort,
         profile.password, profile.method, 600)
     Utils.printToFile(new File(getApplicationInfo.dataDir + "/ss-local-udp-vpn.conf"))(p => {
       p.println(conf)
     })
 
     val cmd = ArrayBuffer[String](getApplicationInfo.nativeLibraryDir + "/libss-local.so", "-V", "-U"
-      , "-b", "127.0.0.1"
+      , "-b", profile.localAddress
       , "-t", "600"
       , "-P", getApplicationInfo.dataDir
       , "-c", getApplicationInfo.dataDir + "/ss-local-udp-vpn.conf")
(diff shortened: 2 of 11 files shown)


DKingCN:
I dont really write scala , so there is another thing I think you may want to know.
The package built from previous commit was installed normal but when I build a release version for this commit , and tried to install it on my angler 7.1.1_r11, the installation process got stuck  then triggered an error to android system (not fatal , system still running good ). But I lost the error log. No more information can be provided. Reboot , all good, ss app is there.
It's running normal so far.

If interested, or out of cautious, you may wanna try to reproduce it?

madeye:
I choose to not merge this change. Because I think It's dangerous to expose the service from a phone, especially on a public Wi-Fi or 3G/4G.

My suggestion is to keep your change as a fork, in case it's useful to anyone.

.gitignore
```diff
@@ -24,6 +24,7 @@ local.sbt
 .idea/
 out/
 .d
+*~
```
Mygod: This line is unneeded. You should put vim/gedit temp file in your global .gitignore or prevent the editor spamming your folder.

src/main/scala/com/github/shadowsocks/ShadowsocksVpnService.scala
```diff
@@ -142,7 +142,7 @@ class ShadowsocksVpnService extends VpnService with BaseService {
       case None => throw NameNotResolvedException()
     }
 
-    handleConnection()
+    handleConnection()///////////////////////////////////////;;;;;;;;;;;;;;;;;;;;;;;;;
```
Mygod: Unneeded comment.

DKingCN:
Thanks to your hints but I dont understand what's the problem of exposing a socket server. Chances to get exploited via socket server can be very low.

LNDDYL:
在POEditor裡面沒有出現。
@Mygod

Mygod:
@LNNDYL Only changes merged into master will be on POEditor.

src/main/scala/com/github/shadowsocks/database/DBHelper.scala
```diff
@@ -88,9 +88,9 @@ class DBHelper(val context: Context)
           profileDao.executeRawNoArgs("ALTER TABLE `profile` RENAME TO `tmp`;")
           TableUtils.createTable(connectionSource, classOf[Profile])
           profileDao.executeRawNoArgs(
-            "INSERT INTO `profile`(id, name, host, localPort, remotePort, password, method, route, proxyApps, bypass," +
+            "INSERT INTO `profile`(id, name, host, localAddress, localPort, remotePort, password, method, route, proxyApps, bypass," +
               " udpdns, auth, ipv6, individual) " +
-            "SELECT id, name, host, localPort, remotePort, password, method, route, 1 - global, bypass, udpdns, auth," +
+            "SELECT id, name, host, localAddress, localPort, remotePort, password, method, route, 1 - global, bypass, udpdns, auth," +
```
Mygod: Just took another look and I think this won't work.

DKingCN: why? this is intended for get local_address  from db and that works. I added a column in db profile.

btw, I am still waiting response from @madeye about the security issue.

Mygod: @DKingCN This isn't how you add columns. See older commit history of this file to learn more about that...

@madeye probably doesn't want this feature available to normal users. If you just want to start a socks5 server, just run a script or something like that.

madeye:
> Chances to get exploited via socket server can be very low.

Actually not. As our `ss-local` works as a SOCKS5 proxy without authentication, an attacker can simply run port scanners to detect the service.

madeye:
As suggested in https://github.com/shadowsocks/shadowsocks-android/pull/1102#issuecomment-275882796, you may keep your changes as a fork. If anyone really needs this feature, they can try your foke instead.

src/main/scala/com/github/shadowsocks/database/DBHelper.scala
DKingCN: XP then sorry i cant help that. I think I can do nothing more than leaving it there. Scala isnt in my plan. Or you got an idea on it?

Mygod: There's nothing Scala specific here (almost plain Java). It's mainly about sqlite and ormlite. I would've helped you fix this if this were being merged though :P.

DKingCN:
@madeye 
Thanks to your reply. I got no problem about fork. Yet I am quite interested about your point on security cause that's what I am majoring in.

So a spammer can simply identify that there is a socket server running on phone but what's the attack surface?
But what can he do with that? I mean I can use nmap on my nethunter to identify services running on a phone and even do ARP poisoning to hijack the traffic, but exposing a proxy doesnt mean more than a spam entry to me. That does not put the phone into a more dangerous situation. (except the traffic fee)
Or we can like easily exploiting a socks5 protocol and hit the kernel now ?

Ohh maybe I got your point, your security means exactly the abuse of socks5 service ?
Cool man , I think I am turning into some solution like a socks5 tunnel on android.

madeye:
@DKingCN A typical usage is botnet for DDOS.

DKingCN:
@madeye 
Different from the bot of pc/IDC server, phone often has a dynamic IP address and attacker cannot maintain the privilege of host like a reverse connection.
Yet that is a considerable situation. thanks.

Mygod:
For further references (since this fork doesn't work yet), one can simply run `/data/data/com.github.shadowsocks/lib/libss-local.so` from a terminal or any app. No root required. However if you pass key via arguments, it can leak to other apps before Android 7.0.

PR #1216, opened 2017-04-14
Add secondary remote dns support.

V-E-O:
### Type of changes

* [ ] Bugfix
* [x] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

Add secondary remote dns support:
192.168.1.1,192.168.1.2
In certain scenario, one may use own custom dns server.
Well, hard code secondary dns 208.67.222.222 may be annoying.
This provides a way to change it.

BTW:
remove duplicated "import java.net.InetAddress" is for "强迫症".

### DIFF
--- a/mobile/src/main/scala/com/github/shadowsocks/BaseService.scala
+++ b/mobile/src/main/scala/com/github/shadowsocks/BaseService.scala
@@ -21,7 +21,6 @@
 package com.github.shadowsocks
 
 import java.io.{File, IOException}
-import java.net.InetAddress
 import java.util
 import java.util.concurrent.TimeUnit
 import java.util.{Timer, TimerTask}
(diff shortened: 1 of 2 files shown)


Mygod:
Secondary DNS isn't the same as alternative DNS in overture afaic. However we can probably support secondary remote DNS.

madeye:
LGTM.

V-E-O:
@Mygod Yes, the alternative dns in overture is usually the untainted dns.
When in bypass chn/gfw list mode alternative dns is user defined remote dns like 8.8.8.8 , but in other modes like global mode, user can only set primary dns and may want to modify this untainted dns which is hard coded with 208.67.222.222.
The result of 208.67.222.222 may be higher priority than user defined primary dns, especially using customized dns (for me, I have customized some dns records conflicting with 208.67.222.222).
This provide a compatible way to implement it without break others.

Mygod:
In that case, perhaps DNS forwarding would work for you?

V-E-O:
My problem can be resolved by disabling AlternativeDNS in ALL/BYPASS_LAN mode.
PrimaryDNS will be enough when you are just need to forwarding the dns, right?
I think overture's OnlyPrimaryDNS config is suitable for this scenario.

And for Secondary DNS, I'm rolling out a patch to set dns list into overture config.
Close this one for deprecated.

PR #1258, opened 2017-05-18
Support for Android O preview

Mygod:
DO NOT MERGE YET! This is a work in progress.

Note: Fixes for NDK compiling issues (and other related stuff) have already been merged to master branch.

### TODO

* [x] Notification channel  
NB: it's possible that NAT service notifications in lock screen are not hiding correctly, needs more tests
* [x] Overture crashing (looks like Android emulator bug)
* [ ] Replace Typeface with native font support with support library (blocked by upstream: https://github.com/scala-android/sbt-android/issues/306)
* [x] Port plugin platform to Android O
* [x] Use PreferenceDataStore to make SharedPreference process-safe (a historical problem)

Lower priorities:

* [x] Adaptive icon?


Mygod:
@madeye Could you investigate why overture is crashing? It seems to crash when launched by Shadowsocks:

```
12067-12756/com.github.shadowsocks:bg D/GuardedProcess: start process: /data/app/com.github.shadowsocks-QkCmGBEGOrqtXp1bS-DAgQ==/lib/x86/liboverture.so -c overture-vpn.conf -V
12067-12756/com.github.shadowsocks:bg W/GuardedProcess: process exit too fast, stop guard: /data/app/com.github.shadowsocks-QkCmGBEGOrqtXp1bS-DAgQ==/lib/x86/liboverture.so -c overture-vpn.conf -V
```

However launching it from `adb shell` (with root/correct uid, and correct cwd) works fine:

```
12858-12858/? I/overture: level=info msg="If you need any help, please visit the project repository: https://github.com/shadowsocks/overture" 
12858-12862/? I/overture: level=info msg="Load IP network file successful" 
12858-12862/? I/overture: level=info msg="Load domain file successful" 
12858-12862/? I/overture: level=info msg="Minimum TTL is 3600" 
12858-12862/? I/overture: level=info msg="CacheSize is 4096" 
12858-12862/? I/overture: level=info msg="Load hosts file successful" 
12858-12862/? I/overture: level=info msg="Start overture on 127.0.0.1:1133" 
```

madeye:
I don't have any Android O device yet. Let me try it on emulator first.

It looks related to new private API protector or something similar.

Mygod:
@madeye I'm just trying it on emulator as well.

Also I just noticed that there's a new option in VPN settings: "Block connections without VPN" for "Always-on VPN". This could be useful for users that are concerned about security.

madeye:
I can reproduce this issue on the emulator now. The process of overture is killed at the very beginning. No related log or dmesg can be found.

madeye:
I just tried 4.1.7. Everything works well on Android O (Pixel and Emulator)...

madeye:
OK, It looks related to https://developer.android.com/preview/features/background.html

Mygod:
@madeye But why does the C binaries work well?

madeye:
I tried starting kcptun plugin with shadowsocks. In that case, ss-local is also killed...

madeye:
Is it possible that the issue is caused by this d5dc3a83793fd03a22a7efeb9cb3ea92d3486da5?

madeye:
Interesting. It seems only the emulator is not working. The latest APK works well on Pixel with Android O.

Mygod:
@madeye Haha what a mysterious bug.

IIRC old GCC binaries (compiled with old ndk) doesn't work either. So no.

You could do a single stepping with gdb. Could be a bug in qemu?

Mygod:
*actually old GCC binaries exhibits more bug even when launched from adb shell (in preview 1). For example, failure to load any file other than overture-vpn.conf. (it gives open : file or directory not found)

Mygod:
It might also relate to some changes in Android for NDK: https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#Writable-and-Executable-Segments-Enforced-for-API-level-26

madeye:
A new finding is it seems we cannot execute a armv7a binary from a arm64-v8a process on Android O. It caused the crash of kcptun plugin and has been fixed via https://github.com/shadowsocks/kcptun-android/commit/90b2b3847aa9211ab49fb4425803055057537eca

Mygod:
That doesn't seem to help.

madeye:
Maybe it's just a bug of x86 emulator? According to my local test, shadowsocks (with kcptun and simple-obfs) works well on a Pixel (Android O).

Let's wait for the next release of Android O SDK and emulator. Hope it will fix our issue.

Mygod:
Okay. Maybe we should file a bug report at Google issue tracker.

Mygod:
@madeye Btw planning on migrating to Kotlin?

madeye:
@Mygod Yes, I think we can schedule this.

madeye:
BTW, is there any tool can easily convert Scala to Kotlin?

Mygod:
Okay I'll look into this and create a branch for Kotlin based off o-preview branch.

Mygod:
@madeye It seems using Kotlin + Gradle is beneficial in the (very) long term and currently there is no automatic tool to convert Scala code to Kotlin code (even for the trivial part) and rewriting this project in Kotlin requires a lot of effort. Let's delay this proposal.

madeye:
@Mygod Understood. I planed to do some porting works during my next vacation, bless I have next vacation... 😞

Mygod:
@madeye Maybe you can write an automatic tool to do that. You might get hired by JetBrains. 😛

I have also written a [basic framework](https://mygod.be/resources/userscloud/7jbmpwa2t2tx). Feel free to build based on this.

Mygod:
@madeye It seems overture is still crashing on the latest emulator. :confused:

madeye:
No function regression is observed in the latest changes.

One weird thing is now we need to long press the number picker dialog to enter the number manually.

madeye:
What about reverting number picker to number only test edit text for port numbers? Actually, I never used the picker to change the port number...

madeye:
A trade off is keeping the number picker for local port, but using edit text for remote port. It helps to avoid illegal local port numbers.

Mygod:
`ShadowsocksRunnerActivity` is not tested. Everything else seems to work fine on Android 7.1.2, including number picker. I will go investigate number picker on Android 8.0 later.

Mygod:
Keyboard is also popping up for me in Android 8.0 emulator.

Mygod:
Adaptive icons are here. [Design source](https://mygod.be/resources/userscloud/etfxe5hhb051).

image

screenshot_1500188693

From: [Adaptive icon playground](https://github.com/nickbutcher/AdaptiveIconPlayground/releases).

Mygod:
screenshot_1500189789

This app is now Android 8.0 ready and this PR is ready to merge.

madeye:
Fix a crash in `ShadowsocksRunnerActivity` that the activity exists too early when preparing VPN in background service.

madeye:
The only one left item is the number picker for porting number.

1. Soft keyboard shows up correctly.
2. However, the number picker is not focused by default.
3. By long pressing the picker, the user is able to input the number manually.

What I've tried:

1. `requestFocus()` at first, but it seems to not work.

madeye:
Surprisingly, after enlarging the text size of number picker, I managed to focus number picker by clicking now...

IMO, It looks that if the number text is too small, the click action may be recognized as scrolling, which prevents focusing the text and enabling manual input.

madeye:
A new crash is observed during auto start:

```
1757 07-19 08:52:48.596  5926  6596 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'void        java.util.concurrent.CountDownLatch.countDown()' on a null object reference
1758 07-19 08:52:48.596  5926  6596 E AndroidRuntime:    at com.evernote.android.job.JobRescheduleService.                          onHandleIntent(JobRescheduleService.java:104)
1759 07-19 08:52:48.596  5926  6596 E AndroidRuntime:    at android.app.IntentService$ServiceHandler.                               handleMessage(IntentService.java:68)
1760 07-19 08:52:48.596  5926  6596 E AndroidRuntime:    at android.os.Handler.dispatchMessage(Handler.java:105)
1761 07-19 08:52:48.596  5926  6596 E AndroidRuntime:    at android.os.Looper.loop(Looper.java:164)
1762 07-19 08:52:48.596  5926  6596 E AndroidRuntime:    at android.os.HandlerThread.run(HandlerThread.java:65)
```

Steps to reproduce this crash:

1. Set auto start enable.
2. Restart the phone.
3. Unlock the phone.
4. Wait until VPNService is started.
5. Open Shadowsocks' main activity.
6. Wait 30 seconds.

It looks an issue of evernote/android-job library.

madeye:
Updating the library to 1.2.0-alpha3 fixed the issue via d8179b5.

Mygod:
Update: It's not possible to dynamically update notification channel's importance: ([source](https://developer.android.com/reference/android/app/NotificationManager.html#createNotificationChannels(java.util.List%3Candroid.app.NotificationChannel%3E)))

> The importance of an existing channel will only be changed if the new importance is lower than the current value and the user has not altered any settings on this channel. All other fields are ignored for channels that already exist.

Mygod:
[Updated design](https://github.com/shadowsocks/shadowsocks-android/files/3292749/ic_launcher.zip) for Google Play new icon specification.

play

PR #1364, opened 2017-09-04
Rewrite network ACLs

Mygod:
### Type of changes

* [x] General refinement

### Details

* A cleaner format;
* Provides better behavior when importing a non-compatible ACL;
* Supports importing URL from a imported file with up to 10 layers.


Mygod:
BREAKING CHANGE: Old network ACLs will be discarded because I removed the parsing part.

PR #1445, opened 2017-11-11
fix fragment called onstart twice on click back

nmyggg:
### fix fragment called onstart twice on click back

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

Phone: HM1S

profilesfragment will onstart twice on call onBackPressed of MainActivity function

my sorry for last commit

### DIFF
--- a/mobile/src/main/scala/com/github/shadowsocks/MainActivity.scala
+++ b/mobile/src/main/scala/com/github/shadowsocks/MainActivity.scala
@@ -243,7 +243,7 @@ class MainActivity extends Activity with ServiceBoundContext with Drawer.OnDrawe
     val tf = Typefaces.get(this, "fonts/Iceland.ttf")
     if (tf != null) title.setTypeface(tf)
 
-    if (savedInstanceState == null) displayFragment(new ProfilesFragment)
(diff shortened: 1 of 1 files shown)


mobile/src/main/scala/com/github/shadowsocks/MainActivity.scala
```diff
@@ -400,7 +417,6 @@ class MainActivity extends Activity with ServiceBoundContext with Drawer.OnDrawe
     val currentFragment = getFragmentManager.findFragmentById(R.id.fragment_holder).asInstanceOf[ToolbarFragment]
     if (!currentFragment.onBackPressed())
       if (currentFragment.isInstanceOf[ProfilesFragment]) super.onBackPressed() else {
-        displayFragment(new ProfilesFragment)
```
Mygod: Seems like this is the only change that's relevant to your title. I will merge this change.

Mygod:
Merged via https://github.com/shadowsocks/shadowsocks-android/commit/86ebbd0db5c281de4d18980521a3fb3d1251ed47. The rest of the changes is reimplemented with https://github.com/shadowsocks/shadowsocks-android/commit/b72fb267e37a03fa375bee8fcc0d36edff2de339.

PR #1446, opened 2017-11-11
Implement Proxy-Only Mode and Transproxy Mode

### LINKED ISSUE #805: Proxy-only mode should be added to work with adblockers and Orbot.
Adblockers like Adguard[2] or AdClear Plus for Android use VPN API to filter ads, which is incompatible with current Shadowsocks-Android. Luckily they have a option to set an upstream SOCKS proxy, we just need an option to run Shadowsocks-Android as a SOCKS proxy listening on localhost to make them compatible, and Adblockers will do the tunneling.

Also Orbot[1](Tor for Android) supports upstream SOCKS proxy, proxy chaining with Shadowsocks can improve performance of Tor greatly as I tested with Shadowsocks client running on my laptop.

Reference:
[1] https://play.google.com/store/apps/details?id=org.torproject.android
[2] https://adguard.com/

### LINKED ISSUE #1441: Manual mode request
### What did you do?

- Use shadowsocks as Orbot's local proxy. 
- Switch proxy through [SwitchyOmega](https://github.com/FelisCatus/SwitchyOmega) in Firefox.

### What did you expect to see?

Shadowsocks-android has an option so I can enter them to Orbot/SwitchyOmega's proxy input box.

### What did you see instead?

- Only VPN and NAT mode supported. If shadowsocks used VPN service, Orbot opens another VPN service, shadowsocks's VPN service will close. If target device isn't rooted, can't use NAT mode.
- Shadowsocks will route all bandwidth, so can't control proxy through other software.

### Additional
[No more “Root” features in Orbot… use Orfox & VPN instead!](https://guardianproject.info/2017/10/27/no-more-root-features-in-orbot-use-orfox-vpn-instead/)

Mygod:
### Type of changes

* [x] General refinement

### Details

With which we finally deprecate NAT mode. Fixes #805. Fixes #1441.

* NAT mode is replaced by transproxy mode;
* `BaseService` classes are moved into `bg` package and refactored in order to minimize duplicate code;
* Local port is moved out of per-profile settings, along with hardcoded DNS port and transproxy (redsocks) port;
* Reset is removed since the only possible clean-up needed can be performed by using force stop in app info;
* Comparing version code before extracting assets is replaced with a more robust approach: comparing update time.

NB: the only code that involves su now is for toggling TFO.

### Possible usecases

Proxy-only mode: Use as an upstream proxy for other proxies.
VPN mode: Mostly unchanged.
Transproxy mode: Meant to be used with AFWall+.

Because I'm lazy, I will call them POM, VPN, TPM from now on.

Here's a way to test POM (and possibly useful too since one could also use it under other modes):

1. Connect to Android using `adb`;
2. Set up port forwarding:
```bash
$ adb forward tcp:<local socks port> tcp:<shadowsocks-android listening port>
```
3. Test out Shadowsocks connection. :smile:

### TODO

Do not merge yet. Need more testing and fixing these TODOs.

* [x] Transproxy script for AFWall+: this could be based on [this script](https://people.torproject.org/~mikeperry/android-hardening/android-firewall/firewall-torify-all.sh). We could put this file somewhere in this project (`firewall-shadowsocks-all.sh` or `scripts/fireweall-shadowsocks-all.sh` etc.). It seems like we could also provide IPv6 support for TPM using `ip6table`.
* [x] After that, we might need to write a guide for using this new mode, like [this tutorial](https://blog.torproject.org/mission-impossible-hardening-android-security-and-privacy);
* [x] Technically we could also implement a basic connection test for other two modes:
  - [x] POM: Just route the traffic through SOCKS5 proxy;
  - [ ] TPM: Do DNS lookup using local DNS and then route the traffic through SOCKS5 proxy (or if possible send it to redsocks?)

@madeye Would you like to work on these? :smile:


madeye:
IMO, connection test is not necessary.
 
I'll test these changes first.

Mygod:
It's good to merge now.

mobile/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,7 @@
                   android:required="false"/>
     <uses-feature android:name="android.software.leanback"
                   android:required="false"/>
-    <uses-feature android:name="android.hardware.camera"
+    <uses-feature android:name="android.hardware.camera"
```
wongsyrone: why space here?

Mygod: https://github.com/shadowsocks/shadowsocks-android/pull/1446/commits/6678fd9541e280963d6680d567c12c4d69e846ef
🤷‍♂️

wongsyrone: @madeye can handle it if he decides to merge

PR #1455, opened 2017-11-17
fix crash when can't find killall command

nmyggg:
### Type of changes

* [x] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

some devices no killall command

### DIFF
--- a/mobile/src/main/scala/com/github/shadowsocks/bg/Executable.scala
+++ b/mobile/src/main/scala/com/github/shadowsocks/bg/Executable.scala
@@ -4,6 +4,9 @@ import java.util.Locale
 import java.util.concurrent.TimeUnit
 
 import android.util.Log
+import android.app.ActivityManager
+import android.content.Context
+import scala.collection.JavaConversions._
 
 /**
   * @author Mygod
@@ -18,10 +21,11 @@ object Executable {
 
   val EXECUTABLES = Array(SS_LOCAL, SS_TUNNEL, PDNSD, REDSOCKS, TUN2SOCKS, OVERTURE)
 
-  def killAll() {
-    val killer = new ProcessBuilder("killall" +: EXECUTABLES: _*).start()
-    if (!killer.waitFor(1, TimeUnit.SECONDS))
-      Log.w("killall", "%s didn't exit within 1s. Post-crash clean-up may have failed."
-        .formatLocal(Locale.ENGLISH, killer.toString))
+  def killAll(context: Context) {
+    val manager = context.getSystemService(Context.ACTIVITY_SERVICE).asInstanceOf[ActivityManager]
+
+    manager.getRunningAppProcesses.foreach(f => if (f.processName.equals(context.getPackageName + ":bg")) {
+      android.os.Process.killProcess(f.pid)
+    })
   }
 }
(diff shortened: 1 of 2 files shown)


Mygod:
Does your ROM not have killall? What's your device?

madeye:
@Mygod I think it's a good fix. There are still many AOSP ROMs don't have `killall`.

mobile/src/main/scala/com/github/shadowsocks/bg/Executable.scala
```diff
@@ -18,10 +21,11 @@ object Executable {
...
-    val killer = new ProcessBuilder("killall" +: EXECUTABLES: _*).start()
-    if (!killer.waitFor(1, TimeUnit.SECONDS))
-      Log.w("killall", "%s didn't exit within 1s. Post-crash clean-up may have failed."
-        .formatLocal(Locale.ENGLISH, killer.toString))
+  def killAll(context: Context) {
+    val manager = context.getSystemService(Context.ACTIVITY_SERVICE).asInstanceOf[ActivityManager]
+
+    manager.getRunningAppProcesses.foreach(f => if (f.processName.equals(context.getPackageName + ":bg")) {
+      android.os.Process.killProcess(f.pid)
+    })
```
Mygod: Could this kill the bg Java process as well? That would be undesirable.

mobile/src/main/scala/com/github/shadowsocks/bg/Executable.scala
```diff
@@ -18,10 +21,11 @@ object Executable {
 
   val EXECUTABLES = Array(SS_LOCAL, SS_TUNNEL, PDNSD, REDSOCKS, TUN2SOCKS, OVERTURE)
 
-  def killAll() {
-    val killer = new ProcessBuilder("killall" +: EXECUTABLES: _*).start()
-    if (!killer.waitFor(1, TimeUnit.SECONDS))
-      Log.w("killall", "%s didn't exit within 1s. Post-crash clean-up may have failed."
-        .formatLocal(Locale.ENGLISH, killer.toString))
+  def killAll(context: Context) {
+    val manager = context.getSystemService(Context.ACTIVITY_SERVICE).asInstanceOf[ActivityManager]
```
Mygod: I don't see a reason to use context here. Just use global app object defined in ShadowsocksApplication.

mobile/src/main/scala/com/github/shadowsocks/bg/Executable.scala
madeye: Yes, we should avoid kill Java processes. Please kill only native processes.

nmyggg: ok it's on your call

madeye:
BTW, I don't think we can list the native processes through Android APIs.

So, maybe it's still a good way to use `killall` here as a "best effort" crashing recovery approach.

Mygod:
I just verified that `killall` does indeed exist on the Android emulator. It seems reasonable to assume that it exists for most devices. However a fallback would still be good. Let me look into this.

Mygod:
Update: `killall` doesn't exist on Android 4.4 emulator.

Mygod:
@madeye Yes `getRunningAppProcesses` only lists Java processes. Closing.

PR #1478, opened 2017-12-03
Porting this project to Gradle + Kotlin

Mygod:
WARNING: This pull request is extremely toxic. You've been warned. :construction: :construction: :construction:

This is one of my stupidest ideas ever.
  
### Type of changes

* [x] Boring changes

### Details

Good things:

* Downloadable fonts!
* Fully integrated with Android Studio with goodies like hybrid debug, layout editor and instant run, etc.
* Apparently sbt-android is almost abandoned anyways.
* It's basically working, at least on my side. (including overture)
* More people will be able to contribute to this project.

Bad things:

* No more multi-inheritance in Kotlin. I wrote a interface + weakhashmap (weak for GC) workaround. Not very readable.
* Number of lines of code got a little bigger (~5%), but still not as annoying as Java. (this is mostly due to the import format in Kotlin)
* ~There has been some weirdness after upgrading, I've fixed some of them but I'm not sure what caused those things;~ (it's my buggy code)
* More people will be able to "pirate" this project.

Note: It's probably a good idea to clone this branch into a new place. :)

FIXME: (assigned to @madeye)

* [ ] Better ways to include native libraries from `core` library (we need this to implement Android TV/Things client, probably not this year. Currently I'm putting some hacks together to make it work)
* [x] Integrate overture building into gradle script; (or a makefile?)
* [x] Travis build;
* [ ] A lot of testing everywhere;
* [ ] More unit tests?


madeye:
Thanks for this **big** patch. 😓 

Let me download a Android Studio first...

madeye:
LGTM.

Merged to https://github.com/shadowsocks/shadowsocks-android/tree/gradle-kotlin.

Mygod:
I think you can push to my branch directly.

madeye:
Got it.

madeye:
It looks still some problems in release build. Maybe related to proguard rules.

madeye:
@Mygod In release build, when accepting server socket in LocalSocketLIstener, we got `EBADF` errors. It looks that that socket object is closed by third-party.

Any thoughts?

Mygod:
Does turning off Proguard help?

madeye:
@Mygod Yes, with proguard disabled, we won't have this issue.

Mygod:
Well I guess it's just a matter of time before we figure out which class Proguard removed that caused this weird behavior then. 😛 Have you tried ``binary search''?

madeye:
Since we can still call `accept()` on serverSocket, it means the object self is not touched but the associate file descriptor is closed by JVM.

Mygod:
I'd be betting on some weird Kotlin method, probably associated with `use`.

madeye:
OK, let me try the ""old"" ways.

madeye:
Someone else encounter the same issue when porting our code to Java...

https://stackoverflow.com/questions/42759762/after-proguard-got-error-when-use-socket-for-cross-process-communication

Mygod:
Have you tried [`useProguard false`](https://github.com/codepath/android_guides/wiki/Configuring-ProGuard)?

Mygod:
Hmmm I reverted 07a48bc20bec32ab427ce67c04ada9a0ddf8550b and f870093c94b212942cb7f132043919016426ca39 and generated a release build and I didn't get EBADF errors. And apk size is under 9MB.

Mygod:
Aha! It seems like the actual problem is a subtle race condition and proguarded Java/Kotlin code runs faster than raw code/(proguarded) scala code. :)

Mygod:
Wait after this patch debug build stops working. :sweat:

madeye:
Please don't revert 07a48bc, which fix another potential problem that localSocket would be GCed because we don't have a valid reference to that object before.

Previously we set the file descriptor to serverSocket, which can be optimized to be a value pass, since It's a int.

I think that's the root cause.

Mygod:
You're right. That's indeed [a problem](https://android.googlesource.com/platform/frameworks/base/+/99a1d45/core/java/android/net/LocalSocketImpl.java#601). Let me fix that later.

Mygod:
I think we can merge this commit into beta branch now. I've already installed c124e1e onto my daily phone.

Mygod:
@madeye That's indeed the root cause. Thanks. :-)

Mygod:
@madeye [This change](https://github.com/Mygod/shadowsocks-android/tree/split-abi) would be able to cut down apk size by almost a half. But you can decide whether to merge it or not. :)

CzBiX:
If the native libraries and VPN services code moved to a library and provided API, I think there will be some new GUI projects based on it.

Mygod:
The plan is actually to use the library to help implement TV UI, etc. I don't think we will ever release the core library available for third-party use.

PR #1491, opened 2017-12-08
Split ABIs for smaller apk size

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

With this change, apk size will go down to <6MB.


madeye:
Let's keep it open. Currently, I don't want to split ABIs.

It's always hard to educate the users which APK they should download.

Mygod:
How about creating two flavors one being universal and the other being split abis?

wongsyrone:
> It's always hard to educate the users which APK they should download.

Agreed.

madeye:
@Mygod It's a good idea.

Mygod:
Unfortunately there doesn't seem to be a way to do this. Settings `split` in variants applies globally.

Mygod:
Update: According to [this](https://stackoverflow.com/a/33230181/2245107), most devices including x86 can run armv7 just fine. We can just simply put armv7 on top.

madeye:
What about the x86 emulator?

Mygod:
I haven't tested yet but probably not since libhoudini is proprietary code. But if somebody wants to play around with emulators, they might as well learn what CPU architectures are already.

Nexus Player uses x86 and [it has libhoudini included since the beginning](https://android.googlesource.com/device/asus/fugu/+/b7e4ded70cd2c71abaa89a16799b29e508781f2d%5E%21/self-extractors/intel/staging/device-partial.mk) and many other Android x86 devices as well.

madeye:
Unless someone would like to run our APK on an AMD device, I think it's safe to remove x86 binaries from our release build.

Mygod:
What about split abis? I still think putting armv7 as the first apk in GitHub releases is a good idea. :)

madeye:
Sure, if there is no issue for x86, I think we can move on to split ABIs.

Mygod:
Conflicts resolved. Please review. @madeye

Mygod:
I just realized there IS a way to make a universal apk (and it's literally called universal apk).

PR #1585, opened 2018-01-19
Min API 19 -> 21

### LINKED ISSUE #1579: This app doesn't support Android 4.3.
Please read FAQ then answer these questions before submitting your issue. Thanks!

### Environment

* Android version: *e.g. "7.0.0_r14", more detailed description is preferred*
* Device: *e.g. Google Pixel XL*
* Shadowsocks version: *version code "v3.0.0" or commit ID "a073f85"*
* Last version that did not exhibit the issue (if applicable):

### Configuration

* [ ] IPv4 server address
* [ ] IPv6 server address
* [ ] Client IPv4 availability
* Encrypt method:
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [ ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [ ] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS Forwarding
* [ ] Auto Connect
* [ ] TCP Fast Open

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

apk size decrease by 50kb yay.

Fix #1579.


wongsyrone:
It's still api 19: https://github.com/shadowsocks/shadowsocks-android/blob/master/core/src/main/jni/Application.mk#L2

Mygod:
Good catch! Fixed.

Mygod:
@madeye Hmmm it seems like that Google Play is erroneously pushing the latest release to Android 4.4 with arm64-v8a devices...

madeye:
@Mygod Why it could happen?

Mygod:
I don't know. It seems like there's no other active apks on play store that has arm64-v8a binaries so that may be causing it. However,

1. I can't seem to find any arm64 device that has Android 4.4 preinstalled; (I didn't try very hard though)
2. Maybe there's a Android 4.4/arm64 specific bug in the package installer?

Mygod:
Google just dropped support for Android 5.1.1 in April's security patch. So it seems like the lifespan for every major release is 3 years. That means only a little more than half of the active devices (as counted by Google Play) is getting the latest security patches.

Personally I'm okay with keeping Lollipop support as the part of code that provides support for API 21 and 22 isn't too bothersome to maintain (unlike 19).

PR #1640, opened 2018-02-23
use kotlin extensions and add snack extension

gengjiawen:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details
* use kotlin extension for android to remove getViewById
* use snack extension
* not fully tested due to go build env in windows

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -4,6 +4,7 @@ import java.util.regex.Pattern
 
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
 
 def getCurrentFlavor() {
     String task = getGradle().getStartParameter().getTaskRequests().toString()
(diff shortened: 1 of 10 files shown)


Mygod:
I'm not a big fan of `kotlin-android-extensions` which is developed by kotlinlang (and I doubt if this is in active development any more). Personally I prefer [data binding library](https://developer.android.com/topic/libraries/data-binding/index.html) to replace `findViewById`.

The method name `View.snack` is possibly confusing. `snackbar` isn't that long to type. And actually I think `Snackbar.make` isn't that terrible since sometimes you need to make additional changes to the snack bar.

In the end, I find nothing worth merging. But thanks for your contribution anyway.

gengjiawen:
Sorry for the test thing, I don't expect this will get reviewed so soon. If kotlin extenstion not a fasion in this repo, and so be it :)

Mygod:
If I don't, @madeye will probably merge it before I know it.

PR #1680, opened 2018-03-15
adjust timeout for ss-local from 600 to 60

Hyzeta:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [ ] General refinement
* Other: Adjust ss-local's timeout from 600 to 60.

### Details

The real timeout for shadowsocks is the minimum of ss-server's timeout and ss-local's timeout. In shadowsocks-android, the origin timeout for ss-local (600) seems to long, so I changed this argument from 600 to 60.

And I think, the best way of handling timeout argument is to let it be decided by the user, not hard-coded in the source file.


madeye:
It's a well tuned number, please don't touch it.

questionyu:
I agree to add one option to let user set a timeout.
(With a larger timeout, GCM can keep a long connection, but not reconnect frequently.)
Just append to global settings.

Issue #1684, opened 2018-03-17
How to build with Gradle?


gam2046:
# Fllowing

- Set environment variable ANDROID_HOME to /path/to/android-sdk
- (optional) Set environment variable ANDROID_NDK_HOME to /path/to/android-ndk (default: $ANDROID_HOME/ndk-bundle)
  * **Not set**
- Set environment variable GOROOT_BOOTSTRAP to /path/to/go
- Fetch submodules using $ git submodule update --init --recursive
- Build it using Android Studio or gradle script

# Code

```
$ git checkout v4.4.5
$ git status
git status
HEAD detached at v4.4.5
```

emmmm, I really want to build with v4.4.5

# Build

```
$ ./gradlew -v
gradlew -v

------------------------------------------------------------
Gradle 4.4.1
------------------------------------------------------------

Build time:   2017-12-20 15:45:23 UTC
Revision:     10ed9dc355dc39f6307cc98fbd8cea314bdd381c

Groovy:       2.4.12
Ant:          Apache Ant(TM) version 1.9.9 compiled on February 2 2017
JVM:          1.8.0_121 (Oracle Corporation 25.121-b13)
OS:           Windows 10 10.0 amd64
```

```
$ go env
set GOARCH=amd64
set GOBIN=
set GOEXE=.exe
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=C:\Users\forDream\go
set GORACE=
set GOROOT=D:\tools\go
set GOTOOLDIR=D:\tools\go\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0
set CXX=g++
set CGO_ENABLED=1
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
```

```
 $ ./gradlew build

gradlew build
Starting a Gradle Daemon, 2 incompatible and 4 stopped Daemons could not be reused, use --status for details
Parallel execution with configuration on demand is an incubating feature.

> Configure project :core 
The CompileOptions.bootClasspath property has been deprecated and is scheduled to be removed in Gradle 5.0. Please use the CompileOptions.bootstrapClasspath property instead.

[...]

Issue #1575, opened 2018-01-15
gradle build failed on win7


xiguapi2015:
Environment

Win7 64bit

What did you do?
I get source and set environment ANDROID_HOME,ANDROID_NDK_HOME ,GOROOT_BOOTSTRAP ,then do "git submodule update --init --recursive".
last,I use "gradle build -stacktrace" then found it fialed as follows:

FAILURE: Build failed with an exception.

* What went wrong:
The specified build file 'E:\work\myproject\workspace\jsq\code\sample_new\shadowsocks-android\uild' does not exist.

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
java.lang.IllegalArgumentException: The specified build file 'E:\work\myproject\workspace\jsq\code\sample_new\shadowsocks-android\uild' does not exist.
        at org.gradle.tooling.internal.provider.StartParamsValidatingActionExecuter.execute(StartParamsValidatingActionExecuter.java:43)
        at org.gradle.tooling.internal.provider.StartParamsValidatingActionExecuter.execute(StartParamsValidatingActionExecuter.java:30)
        at org.gradle.tooling.internal.provider.SessionFailureReportingActionExecuter.execute(SessionFailureReportingActionExecuter.java:59)
        at org.gradle.tooling.internal.provider.SessionFailureReportingActionExecuter.execute(SessionFailureReportingActionExecuter.java:44)
        at org.gradle.tooling.internal.provider.SetupLoggingActionExecuter.execute(SetupLoggingActionExecuter.java:45)
        at org.gradle.tooling.internal.provider.SetupLoggingActionExecuter.execute(SetupLoggingActionExecuter.java:30)
        at org.gradle.launcher.daemon.server.exec.ExecuteBuild.doBuild(ExecuteBuild.java:67)
        at org.gradle.launcher.daemon.server.exec.BuildCommandOnly.execute(BuildCommandOnly.java:36)
        at org.gradle.launcher.daemon.server.api.DaemonCommandExecution.proceed(DaemonCommandExecution.java:122)
        at org.gradle.launcher.daemon.server.exec.WatchForDisconnection.execute(WatchForDisconnection.java:37)
        at org.gradle.launcher.daemon.server.api.DaemonCommandExecution.proceed(DaemonCommandExecution.java:122)
        at org.gradle.launcher.daemon.server.exec.ResetDeprecationLogger.execute(ResetDeprecationLogger.java:26)
[...]

PR #1696, opened 2018-03-22
Fix building with Windows. #1684

gam2046:
Signed-off-by: forDream <gam2046@users.noreply.github.com>

### Type of changes

* [x] Bugfix
* [ ] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

- Supports building projects on windows.
- Python is required

### DIFF
--- a/core/src/overture/make.bat
+++ b/core/src/overture/make.bat
@@ -28,7 +28,7 @@ MKDIR %TARGET%\armeabi-v7a>nul 2>nul
 MKDIR %TARGET%\x86>nul 2>nul 
 MKDIR %TARGET%\arm64-v8a>nul 2>nul 
 
-SET CC=%ANDROID_NDK_HOME%\toolchains\arm-linux-androideabi-4.9\prebuilt\windows-x86_64\bin\arm-linux-androideabi-gcc.exe
+SET CC=%ANDROID_ARM_TOOLCHAIN%\bin\arm-linux-androideabi-gcc.exe
 
 REM Check environment availability
 IF NOT EXIST %CC% (
--- a/core/src/overture/make.bat
+++ b/core/src/overture/make.bat
@@ -28,21 +28,35 @@ MKDIR %TARGET%\armeabi-v7a>nul 2>nul
 MKDIR %TARGET%\x86>nul 2>nul 
 MKDIR %TARGET%\arm64-v8a>nul 2>nul 
 
+SET CC=%ANDROID_NDK_HOME%\toolchains\arm-linux-androideabi-4.9\prebuilt\windows-x86_64\bin\arm-linux-androideabi-gcc.exe
+
+REM Check environment availability
+IF NOT EXIST %CC% (
+    ECHO "gcc not found"
+    EXIT 1
+)
+
+WHERE python.exe
+IF "%ERRORLEVEL%" == 1 (
+    ECHO "python not found"
+    EXIT 1
+)
+
 IF NOT EXIST %ANDROID_ARM_CC% (
 	ECHO "Make standalone toolchain for ARM arch"
-    %ANDROID_NDK_HOME%\build\tools\make_standalone_toolchain.py --arch arm ^
+    python.exe %ANDROID_NDK_HOME%\build\tools\make_standalone_toolchain.py --arch arm ^
         --api %MIN_API% --install-dir %ANDROID_ARM_TOOLCHAIN%
 )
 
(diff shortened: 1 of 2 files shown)


core/src/overture/make.bat
```diff
@@ -28,21 +28,35 @@ MKDIR %TARGET%\armeabi-v7a>nul 2>nul
 MKDIR %TARGET%\x86>nul 2>nul 
 MKDIR %TARGET%\arm64-v8a>nul 2>nul 
 
+SET CC=%ANDROID_NDK_HOME%\toolchains\arm-linux-androideabi-4.9\prebuilt\windows-x86_64\bin\arm-linux-androideabi-gcc.exe
```
wongsyrone: It seems you are using the outdated ndk

gam2046: In fact, the ndk I used is not outdated.
But I found that maybe could be written like this:
`SET CC=%ANDROID_ARM_TOOLCHAIN%\bin\arm-linux-androideabi-gcc.exe`

madeye: I think this one is better. Please update the path here.

wongsyrone:
CC variable is set below, why we need this at the very beginning?

wongsyrone:
If you want to check the availability of `ANDROID_{ARCH}_CC`, I don't think this makes any sense, the toolchain installer makes sure it does exist.

And it seems  `WHERE python.exe` doesn't work on my box.

gam2046:
`goBuild' requires gcc to compile(I'm not good at Golang, maybe gcc is needed in Golang), and there are no gcc environment variables by default for windows, so `CC` variables need to be set. So that the Golang compiler knows the location of the gcc.

I am sure that the `where` commands is available on Windows 7/10, and I compile it on Windows 10

wongsyrone:
New NDK deprecates GCC and CC is set at https://github.com/gam2046/shadowsocks-android/blob/207d76a143a4e9bb9b0e1a58a7f122060aae63c2/core/src/overture/make.bat#L95

gam2046:
https://github.com/shadowsocks/shadowsocks-android/blob/master/core/src/overture/make.bat#L53
https://github.com/shadowsocks/go/blob/master/src/make.bat#L29

Maybe gcc needs to be used in golang?

Now you define `CC` on line 95, and you need to use `CC` on line 53

wongsyrone:
It's host compiler, not the cross-compiler. Since widely used Linux distro has gcc installed by default. That env won't cause problems. For Windows, you have to choose a host compiler to compile *.c src correctly. As far as I know, Go's bootstrap doesn't rely on C now.

gam2046:
Is this related to the version of GO? 
At compile time, I use go 1.7 to indicate that gcc cannot be found.
Finally, I added gcc to allow the compilation to continue.

wongsyrone:
Not sure about that, I don't use windows to compile this project. BTW, you want to use https://github.com/shadowsocks/go/tree/shadowsocks-go1.10, it's golang 1.10.

gam2046:
emmmm I'm not good at `GO`. I don't know the relationship between these versions :
- Version of https://github.com/shadowsocks/go/
- Version of Go binary : https://golang.org/dl/
- Version of Go's bootstrap

I tried to compile this project in Linux, no problem.  But try to compile in Windows, encounter some problems, and try to fix these problems, one of the problems is that can't find gcc

Mygod:
No additional comments from me. CC @imReker

imReker:
I create windows build.bat in less than 5 minutes, just simply translate sh to bat, so it doesn't have error check, same as sh version. It just do right thing when everything is OK.

Actually the build.bat can't work on most people's machine because of the Build of GoLang, like #1575.
GCC for windows is required, and my solution is a Windows 10 machine and build GoLang within WSL, then build.bat will success. 

@wongsyrone 
`WHERE python.exe` will not work if python is not in %PATH%.
This check is OK because user will see an error message in console instead of a Open With window when execute py file.

wongsyrone:
Then my opinion is building on windows doesn't worth fixing, there are so many issues ordinary users cannot handle. Just let who wants to compile use *nix.

imReker:
Agree. Waste time on Windows build is not worth. I just make Windows build possible on some machine, but not perfect on every machine. If someone really want to build on Windows, read build.bat and related pr comment is enough.
Most people just need overture.so bin to finish the build progress.

So, add windows build.bat to `task goBuild` maybe is a bad idea. Remove it and people will think it can not be built on Windows.

PR #1698, opened 2018-03-23
Actually use underlying network

### LINKED ISSUE #1666: App do not behave normally using vpn via wifi
Please read FAQ then answer these questions before submitting your issue. Thanks!

### Environment

* Android version: 8.1.0 OPM1.17.10.19.013
* Device: OnePlus 3T
* Shadowsocks version: 4.5.1
* Last version that did not exhibit the issue (if applicable): Android 8.0 and before

### Configuration

* [x] IPv4 server address
* [ ] IPv6 server address
* [x] Client IPv4 availability
* Encrypt method: rc4
* Route
  * [x] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [ ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [ ] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS Forwarding
* [x] Auto Connect
* [x] TCP Fast Open

### What did you do?
Connecting the VPN using wifi network and use Google Photos

### What did you expect to see?
Google Photos back up automatically.

### What did you see instead?
Google Photos does not backup at all

Android Issue Tracker: https://issuetracker.google.com/issues/68657525
And latest reply from Googler:

> Status: Won't Fix (Intended Behavior)
When a VPN is up the OS has no way to guarantee that network traffic will go on wifi. The VPN app could decide to send and receive traffic on any available interface, including cell data. Recent changes to the VPN code have made it so that if the VPN calls setUnderlyingNetworks, the networking APIs will reflect some of the properties of the underlying networks (e.g., whether they are metered or not) into the VPN's networkcapabilities. But it's up to the podcast app to use the required APIs, such as isActiveNetworkMetered.

Not sure this is a Shadowsocks client issue, but since Google Play can not download with vpn via wifi. I think it might be a shadowsocks issue.

Thanks

Mygod:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

This pull request uses `setUnderlyingNetwork` not only as a hint, but also uses the underlying network it sets for VPN connections.

Everything seems to be working fine. Might need more testing.

This also fixes #1666.

### DIFF
--- a/mobile/src/main/AndroidManifest.xml
+++ b/mobile/src/main/AndroidManifest.xml
@@ -3,6 +3,7 @@
           package="com.github.shadowsocks"
           android:installLocation="internalOnly">
 
+    <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.NFC" />
--- a/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
@@ -20,7 +20,6 @@
 
 package com.github.shadowsocks.bg
 
-import android.annotation.TargetApi
 import android.app.Service
 import android.content.Context
 import android.content.Intent
@@ -29,7 +28,6 @@ import android.net.*
 import android.os.Build
 import android.os.IBinder
 import android.os.ParcelFileDescriptor
-import android.support.v4.os.BuildCompat
 import android.util.Log
 import com.github.shadowsocks.App.Companion.app
 import com.github.shadowsocks.JniHelper
@@ -56,7 +54,8 @@ class VpnService : BaseVpnService(), LocalDnsService.Interface {
         private val getInt: Method = FileDescriptor::class.java.getDeclaredMethod("getInt$")
 
         /**
-         * Unfortunately registerDefaultNetworkCallback is going to return our VPN interface: https://android.googlesource.com/platform/frameworks/base/+/dda156ab0c5d66ad82bdcf76cda07cbc0a9c8a2e
+         * Unfortunately registerDefaultNetworkCallback is going to return VPN interface since Android P DP1:
+         * https://android.googlesource.com/platform/frameworks/base/+/dda156ab0c5d66ad82bdcf76cda07cbc0a9c8a2e
          *
          * This makes doing a requestNetwork with REQUEST necessary so that we don't get ALL possible networks that
          * satisfies default network capabilities but only THE default network. Unfortunately we need to have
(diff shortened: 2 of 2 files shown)


madeye:
On second thought, I think maybe this is not necessary. As Linux OS would take care of the default underlying network very well, this change may introduce issues on many broken third-party AOSP.

Mygod:
Welp. I never said this is necessary. I even said `might need more testing'. :/

Mygod:
And actually it's Android system that's taking care of the default networks, at least since 5.0 (performed by Netd and other components). You can see in the source code that `VpnService.protect` is actually just binding the socket to the default network in `Netd`.

This is only meant to be a somewhat better fix for #1666 (than simply reporting what the default network is as I did [in beta branch](https://github.com/shadowsocks/shadowsocks-android/commit/3a8f863bd1f0deed081de082a674020368f30319)).

madeye:
Hmm, I thought it's a must-have for #1666.

Let's revert it for now. I think the fix in beta branch is more reasonable.

Mygod:
I think this is the more reasonable thing to do though. What do you say if we use this behavior only for Android P and up?

madeye:
Yes, it's doable.

I think #1666 is actually an issue of Google, what we need to do is to get ready for Google's fix in Android P.

Mygod:
@madeye By the way #1706 is caused by `requestNetwork` which is also used in beta branch (because `addOnDefaultNetworkActiveListener` doesn't work). So far I haven't seen the harm of doing `Network.bindSocket` instead of `protect`.

madeye:
@Mygod Interesting, If `bindSocket` works, maybe we don't need to bind socket in Java.

Do you think if we can do socket binding in Linux way like this?

```
struct ifreq ifr;

memset(&ifr, 0, sizeof(ifr));
snprintf(ifr.ifr_name, sizeof(ifr.ifr_name), "eth0");
if (setsockopt(s, SOL_SOCKET, SO_BINDTODEVICE, (void *)&ifr, sizeof(ifr)) < 0) {
    ... error handling ...
}
```

madeye:
Oops, `SO_BINDTODEVICE` requires ROOT permission.

Mygod:
@madeye Yeah you don't think it would be that easy huh? :smile: 

Android has lots of permissions enforcements and what `protect` actually does is it sends the fd to `ConnectivityService` (which checks that caller has the permissions) iirc which then passes to `Netd` (who knows the default network) to do the actual binding.

Mygod:
Personally I think what we have right now is the best way of doing it and I'd like to keep it this way if it doesn't give rise to other issues for some devices.

Mygod:
@madeye Some related source code: https://android.googlesource.com/platform/system/netd/+/1564d48/client/NetdClient.cpp#227

From the analytics, I can see there are ~44k exceptions thrown by `bindSocket` in 4.5.2 across Android 6.0 - 8.0 (interestingly not for 8.1 yet). Unfortunately there's not enough information to tell what's going on. There's also nothing to compare to because we don't track when `protect` returns false. We should also expect this to throw exceptions because there's an unavoidable race condition when network switch happens (also the case for `protect`, you just can't stop an interface from going down, e.g. when a user is unplugging the Ethernet cable).

There's also a downside of using `bindSocket`, it will internally throw an exception to make sure the socket is not connected. This might have a negative impact on battery life and performance. See source code:

```
    public void bindSocket(FileDescriptor fd) throws IOException {
        try {
            final SocketAddress peer = Os.getpeername(fd);
            final InetAddress inetPeer = ((InetSocketAddress) peer).getAddress();
            if (!inetPeer.isAnyLocalAddress()) {
                // Apparently, the kernel doesn't update a connected UDP socket's
                // routing upon mark changes.
                throw new SocketException("Socket is connected");
            }
        } catch (ErrnoException e) {
            // getpeername() failed.
            if (e.errno != OsConstants.ENOTCONN) {
                throw e.rethrowAsSocketException();
            }
        } catch (ClassCastException e) {
            // Wasn't an InetSocketAddress.
            throw new SocketException("Only AF_INET/AF_INET6 sockets supported");
        }

        final int err = NetworkUtils.bindSocketToNetwork(fd.getInt$(), netId);
        if (err != 0) {
            // bindSocketToNetwork returns negative errno.
            throw new ErrnoException("Binding socket to network " + netId, -err)
                    .rethrowAsSocketException();
        }
    }
```

If possible, I'd love to re-enable `bindSocket` for Android 7.0-8.1 as well to see which one works better.

PR #1733, opened 2018-04-05
Update translation of FAQ

initrider:
* [ ] Bugfix
* [ ] New feature
* [x] Translation
* [ ] General refinement


amayatsky:
@initrider Looks like Yota is still shaping the Shadowsocks traffic. Maybe that's because a non-standard port is used. Any update on that?

Issue #1740, opened 2018-04-09
4.5.5出现DNS错误


triaqu:
Please read FAQ then answer these questions before submitting your issue. Thanks!

### Environment

* Android version: 8.1
* Device: Oneplus 5
* Shadowsocks version: 4.5.5 服务器端为最新版本
* Last version that did not exhibit the issue (if applicable):
4.5.3.
### Configuration

_Put an `x` inside the [ ] that applies._

* [ x] IPv4 server address
* [ ] IPv6 server address
* [x ] Client IPv4 availability
* [ ] Client IPv6 availability
* Encrypt method: aes-128-gcm
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [x ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [x ] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [x ] DNS Forwarding
* Plugin configuration (if applicable):
* [x ] Auto Connect
* [ ] TCP Fast Open
* If you're not using VPN mode, please supply more details here:

### What did you do?
如果任意连接一个错误的配置后，切换到其他正常的配置。

### What did you expect to see?
能正常使用。

### What did you see instead?
切换到正常配置后也无网络连接，无论是否使用远程DNS是否使用obfs, 都会出现无法解析谷歌地址的错误。

开关不起作用，需要杀掉进程后重启才能使用。

madeye:
I cannot reproduce your issue locally.

It looks a known issue of Oneplus' latest ROM that it wouldn't flush DNS cache when VPN connection established or disconnected. https://github.com/shadowsocks/shadowsocks-android/issues/1721#issuecomment-379121367

triaqu:
@madeye  谢谢解答，但有个疑问，4.5.3版本确可以使用正常使用。

Love4Taylor:
@madeye 4.5.5, Google Pixel, Android P 有同样的问题, 当某一个配置文件连不上之后不管换成哪个都是无速度无响应, 杀进程后才可恢复

madeye:
@Love4Taylor Submit the output of `adb shell ps` and `adb logcat -s shadowsocks` here, after you reproduce the issue.

Mygod:
@madeye Interestingly, I noticed a similar bug as this one with #1742. Here's related log:

```
[...]

PR #1742, opened 2018-04-09
Get rid of -V flag

Mygod:
:construction: **ONLY INTENDED FOR DEBUGGING PURPOSES DO NOT MERGE** :construction:

By disallowing this app itself, every socket created by this app will go through default network automatically. However, since it's called *disallowing*, Android system won't let us bind to VPN tunnel. Therefore we will have to resort to using proxies to connect with Shadowsocks.

Pros:
1. No more ancillary fd passing;
2. ~No more go lib patching;~
3. Easier plugin development.

Cons:
1. ~Ads might not work if Google is blocked; (unless it goes through Google Service Framework)~
2. Connection test will have to test through proxy instead of the better testing method through VPN tunnel;
3. All sockets will use default network and therefore we are no longer using underlying network;
4. ~Only works on Android 5.0+.~

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
@@ -159,6 +159,8 @@ class VpnService : BaseVpnService(), LocalDnsService.Interface {
                         }
                     }
             if (profile.bypass) builder.addDisallowedApplication(me)
+        } else {
+            builder.addDisallowedApplication(packageName)
         }
 
         when (profile.route) {
--- a/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
@@ -26,7 +26,6 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.net.*
-import android.os.Build
 import android.os.IBinder
 import android.os.ParcelFileDescriptor
 import android.support.v4.os.BuildCompat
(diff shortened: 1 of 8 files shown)


madeye:
I planed to do this years ago, but it wouldn't work if we need to support Android 4.x at that time.

One concern is about the plugin mode. Would it also work with SIP003 plugin?

Mygod:
@madeye Yes since we are running plugins under our UID.

Mygod:
Disclaimer: I actually haven't tested it with plugins. But theoretically it should. But I've had enough debugging for a day.

madeye:
Never mind, I submitted a fix for it.

mobile/src/main/java/com/github/shadowsocks/bg/VpnService.kt
```diff
@@ -159,6 +159,8 @@ class VpnService : BaseVpnService(), LocalDnsService.Interface {
                         }
                     }
             if (profile.bypass) builder.addDisallowedApplication(me)
+        } else {
+            builder.addDisallowedApplication(packageName)
```
Mygod: After adding these two lines, I can no longer reproduce #1740. Good catch though!

Mygod:
After thinking more about this, I think this is not the way Android VpnService is intended to be used, so even though this might work, it's still more of a hack. I think we could keep this open for plugin developers as a debugger tool?

madeye:
Sure. Let's keep it open in case someone else may find it useful.

ayanamist:
Since golang/go#9661 is fixed, "go lib patching" is no more necessary in upcoming Go 1.11.

PR #1747, opened 2018-04-10
Manage all processes with GuardedProcessPool

Mygod:
### Type of changes

* [x] General refinement

### Details

This way we can destroy the processes and wait for them to exit in parallel, and ensure the process is destroyed so that we won't encounter strange port being used issue at the same time.

This addresses #1740 but probably doesn't fix it.

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/bg/BaseService.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/bg/BaseService.kt
@@ -76,7 +76,7 @@ object BaseService {
         @Volatile var state = STOPPED
         @Volatile var plugin = PluginOptions()
         @Volatile var pluginPath: String? = null
-        var sslocalProcess: GuardedProcess? = null
+        val processes = GuardedProcessPool()
 
         var timer: Timer? = null
         var trafficMonitorThread: TrafficMonitorThread? = null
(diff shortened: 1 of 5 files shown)


madeye:
I think we need sigkill after waitfor timeout. Otherwise, it may not fix #1740.

Mygod:
`Process.waitFor` requires API 26. And I'd also rather not use `sigkill` unless necessary. Could you investigate what would cause `ss-local` to ignore `sigterm`?

madeye:
Actually, I cannot report the issue locally.

According to the report:

> 1。断开正常连接，切换到错误配置，连接报错，关闭连接。切换到正常配置，连接报错。
> 2，断开正常连接，切换到错误配置，连接报错，不关闭直接切换到正常配置，连接正常。

Mygod:
Yes. Me neither. But there does seem to be some edge cases where `libss-local` doesn't respond to `SIGTERM`. For example, in #1742 when the socket is accidentally not protected, or the new example I just posted at #1740.

madeye:
After instrumenting ss-local, my observation is that the ss-local process hangs when `stop_plugin()`. It means it handles SIGTERM correctly, but the child process didn't.

ss-local calls cork_subprocess_group_abort then waitpid until the child process finished.

https://github.com/shadowsocks/libcork/blob/c9b3ea9638b3adcbc07ee229fd630f439cdecae8/src/libcork/posix/subprocess.c#L483

Mygod:
@madeye Okay. Here's the source code: https://github.com/shadowsocks/kcptun-android/pull/32

PR #1751, opened 2018-04-12
fix unit

onlymash:
### Type of changes

* [x] Bugfix
* [ ] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

https://en.wikipedia.org/wiki/Binary_prefix

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/bg/TrafficMonitor.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/bg/TrafficMonitor.kt
@@ -41,7 +41,7 @@ object TrafficMonitor {
     @Volatile
     private var dirty = true
 
-    private val units = arrayOf("KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB", "BB", "NB", "DB", "CB")
+    private val units = arrayOf("KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB", "BiB", "NiB", "DiB", "CiB")
     private val numberFormat = DecimalFormat("@@@")
     fun formatTraffic(size: Long): String {
         var n = size.toDouble()


Mygod:
No need for this. There's no ambiguity here.

onlymash:
这里的计算方法是除以1024，是二进制千啊。如果你要用十进制，就应该改成1000，并且把K改成小写。
```
                           SI     BINARY

                   0:        0 B        0 B
                  27:       27 B       27 B
                 999:      999 B      999 B
                1000:     1.0 kB     1000 B
                1023:     1.0 kB     1023 B
                1024:     1.0 kB    1.0 KiB
                1728:     1.7 kB    1.7 KiB
              110592:   110.6 kB  108.0 KiB
             7077888:     7.1 MB    6.8 MiB
           452984832:   453.0 MB  432.0 MiB
         28991029248:    29.0 GB   27.0 GiB
       1855425871872:     1.9 TB    1.7 TiB
 9223372036854775807:     9.2 EB    8.0 EiB   (Long.MAX_VALUE)
```

Mygod:
Please convince Android team to fix their unit issues first. We have to keep consistency with the system.

onlymash:
那好，来看看 Android Oreo 用什么单位。
https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/text/format/Formatter.java

/**
     * Formats a content size to be in the form of bytes, kilobytes, megabytes, etc.
     *
     * <p>As of O, the prefixes are used in their standard meanings in the SI system, so kB = 1000
     * bytes, MB = 1,000,000 bytes, etc.</p>
     *
     * <p class="note">In {@link android.os.Build.VERSION_CODES#N} and earlier, powers of 1024 are
     * used instead, with KB = 1024 bytes, MB = 1,048,576 bytes, etc.</p>
     *
     * <p>If the context has a right-to-left locale, the returned string is wrapped in bidi
     * formatting characters to make sure it's displayed correctly if inserted inside a
     * right-to-left string. (This is useful in cases where the unit strings, like "MB", are
     * left-to-right, but the locale is right-to-left.)</p>
     *
     * @param context Context to use to load the localized units
     * @param sizeBytes size value to be formatted, in bytes
     * @return formatted string with the number
     */

下图是系统自带的流量统计 0.5MB == 500kB
screenshot_settings_20180417-103217 01
screenshot_settings_20180417-103213 01

Mygod:
OK. Let's use `Formatter.formatFileSize` instead.

Mygod:
Fixed via 98df31c25819abb0be64f9d9423dd7e187d0bcda.

onlymash:
直接用系统单位可能有的人无法接受，因为在中文环境， 8.1 把单位也翻译了。。

Mygod:
They should file a bug report to Android then. I'm always happy to keep consistency with the system and/or write less code.

PR #1762, opened 2018-04-18
make connection test url configurable

V-E-O:
### Type of changes

* [ ] Bugfix
* [x] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

When SS is used in China as a pure secure proxy, the hardcoded connection test Google server may not work.
Here I added two global URL configuration, one for China List mode, one for else.

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/MainActivity.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/MainActivity.kt
@@ -166,9 +166,7 @@ class MainActivity : AppCompatActivity(), ShadowsocksConnection.Interface, Drawe
      */
     private fun testConnection(id: Int) {
         val (success, result) = try {
-            val urlString = when (app.currentProfile!!.route) {
-                Acl.CHINALIST -> DataStore.testUrlChinaList
-                else -> DataStore.testUrl}
+            val urlString = DataStore.testUrl
             val url = URL(urlString)
             if (url.protocol != "https")
                 throw MalformedURLException(getString(R.string.connection_test_url_non_https, urlString))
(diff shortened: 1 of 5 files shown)


V-E-O:
Oops, Codacy reported the ConstantNaming issue (uppercase), seems that we need some unification for the original code here.

madeye:
Please don't add two URLs, One test URL is already enough.

V-E-O:
@madeye My first version is one URL, but after my concern is that it may break original design (ChinaList mode using qualcomm.cn).
You guys make decision, I will pull the one URL version later.

Mygod:
Wait what? I don't see this change necessary.

V-E-O:
@Mygod I believe it's usefull when you just use SS as a secure proxy within China's internet, because hardcoded www.google.com/generate_204 breaks the connectivity test feature.
Also my first patch (two configrable url) does not "kills off" any "feature".

Mygod:
This app isn't going to cover every single use case ever.

Mygod:
Connectivity test is never meant to be a core feature - you don't see stuff like this in other clients. You could always use your browser to test if stuff works. If I merge changes like this, the next thing you know is that you'll have hundreds of (mostly useless) configuration options in this app. Additional rule for ChinaList mode is there since it doesn't complicates the UI too much.

But feel free to make forks if you think this is useful to you.

V-E-O:
Other user scenario like this one https://github.com/shadowsocks/shadowsocks-android/issues/1603
If your server is built on ADSL with dyn IP, broken connectivity test is painful.

V-E-O:
I've a fork version, but to maintain and build is a heavy stuff.
Suppose you don't want UI change, what if check both two URL (Google and Qualcomm) in the backend?

PR #1768, opened 2018-04-21
Fix windows building issues

ewrfdv:
### Type of changes

* [x ] Bugfix
* [ ] New feature
* [ ] Translation
* [ ] General refinement
* Other:

### Details

Install toolchains in core/src/overture/.deps before checking gcc.


Mygod:
Why would you want to check for toolchain twice?

ewrfdv:
Well, the second check is unnecessary，now I corrected it.

PR #1771, opened 2018-04-23
add connection test backup plan

V-E-O:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

As discussed in https://github.com/shadowsocks/shadowsocks-android/pull/1762, here I proposed a solution not changing UI.
It's not a core feature change, thus this enhancement does not affect the original function. Why non-core feature change like UI acceptable but this one? At least this change may relieve lots of network diagnostic work for many.

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/MainActivity.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/MainActivity.kt
@@ -93,6 +93,7 @@ class MainActivity : AppCompatActivity(), ShadowsocksConnection.Interface, Drawe
     private var previousSelectedDrawer: Long = 0    // it's actually lateinit
 
     private var testCount = 0
+    private var testPhase = -1
     private lateinit var statusText: TextView
     private lateinit var txText: TextView
     private lateinit var rxText: TextView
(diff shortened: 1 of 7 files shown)


madeye:
Unfortunately, I don't think this approach makes any sense.

It's important to use google.com for routing modes rather than ChinaList, since we want to make sure there is no DNS poisoning.

madeye:
@Mygod Any thoughts?

Mygod:
How about just no?

madeye:
@V-E-O I think we have to refuse this change. The major reason is that we don't want to confuse our users (most of them even don't know what a test URL is).

As @Mygod mentioned, you can keep your changes in a fork. 

Anyway, thanks for your contribution!

V-E-O:
Why no? It's a bug. Bug in non-core module, useless? ChinaList test URL(qualcomm) was resolved well in this non-core module.

I think it's better to add a note "you're not welcomed if your SS server inside a place Google inaccessible".

Mygod:
Yes I can see it's useful to you. But I'd say there are at most a few dozens of people who would find this useful, and these people probably would also have enough skills (DDNS? really?) to merge the fork and compile this repo anyways (but that's not the point). And in this sense, it's useless.

I can probably come up with 100 features like this that may be useful to somebody sometime. If we support all that, we will end up designing Microsoft Word, but open source.

> It's not a bug, it's a feature.

Btw I also agree that this is hella confusing to other users.

V-E-O:
> Yes I can see it's useful to you.

This "you" is not "me" but "us", like https://github.com/shadowsocks/shadowsocks-android/issues/1603

> But I'd say there are at most a few dozens of people who would find this useful, and these people 
probably would also have enough skills (DDNS? really?) to merge the fork and compile this repo anyways (but that's not the point).

Everyone needs to rebuild a bug fix version himself/herself, really?

> And in this sense, it's useless.

A few dozens of people may use ChinaList mode, then test URL (qualcomm) is useless by your logic. Better revert it?

> I can probably come up with 100 features like this that may be useful to somebody sometime. If we support all that, we will end up designing Microsoft Word, but open source.

It's not a feature but a bug. There's a patch to fix this, that's the advantage of open source. I once reported a crash of Apple's product, they would not like to fix it, but open source usually do.

> Btw I also agree that this is hella confusing to other users.

The first manual configuration design is not confusing at all.

> EDIT: Also you do know that you can use your browser to test connectivity as well, right?

That's an excellent excuse for not fixing the bug, like saying, you know you can use other secure proxy to replace SS right?

Mygod:
You're completely missing all the points.

> Just no.

PR #1794, opened 2018-05-09
Migrating to Room

### LINKED ISSUE #1715: Race conditions in OrmLite
Issues like https://github.com/j256/ormlite-core/issues/139 causes `SQLiteConstraintException` and thus crash (although rare).

We might need to migrate to other ORM libraries. [Room 1.1.0](https://developer.android.com/topic/libraries/architecture/room.html) looks good.

Contributions are welcome.

Mygod:
### Type of changes

* [x] Bugfix

### Details

Fixes #1715. In comparison, Room library enables better control over SQLite queries, and seems to be actually more lightweight than OrmLite (currently -56 KB in apk size).

TODO:

* [ ] Rewrite migrations for old database versions (private < 25, public < 2).
* [ ] Add tests for migrations?

Changes:

* `Profile.id` is now `long` to keep consistency with SQLite;
* `Profile.date` is removed since I realized it's persisted as `VARCHARS` in old versions.


madeye:
I think we don't need to migrate database for "old" versions (Scala ones). That should make life easier.

Mygod:
Right now it migrates from the latest version of the database so you're suggesting only adding support for versions since it's rewritten in Kotlin?

Mygod:
Okay... I suggest warning users that data loss could happen if they're upgrading from a version that's too old.

PR #1806, opened 2018-05-19
Add distributionSha256Sum property to gradle-wrapper.properties

Rudloff:
### Type of changes

* [ ] Bugfix
* [ ] New feature
* [ ] Translation
* [x] General refinement
* Other:

### Details

This makes Gradle check the binary it downloads.


Mygod:
Source?

Mygod:
Thanks but `graddle-wrapper.properties` in this project is generated by Android Studio. Maybe you'd like to report this issue to them instead.

Mygod:
For now I don't think it's meaningful to check integrity of gradle but not other plugins. Security wise https is already used.

PR #1878, opened 2018-07-15
Replace Fragment with Controller

onlymash:
https://github.com/bluelinelabs/Conductor


Mygod:
I can't see any benefits either. And why should we use this library but not others, or nothing at all? Furthermore, the library doesn't even seem to be in active development.

PR #1881, opened 2018-07-16
Migrate to Android X

Mygod:


Mygod:
Oh and also there's this bug: https://issuetracker.google.com/issues/64729576

onlymash:
既然都上 androidx 了，要不要用 Navigation Architecture Component?

Mygod:
We're using Android X because support library is being deprecated.

I don't see why we should use navigation architecture component. What's the advantage of using that?

PR #1894, opened 2018-07-27
speed up for gradle

pexcn:
- Enable android build cache
- Increase jvmargs to 2G

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,4 +17,4 @@ org.gradle.jvmargs=-Xmx2048M
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-org.gradle.parallel=true
+# org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,10 +9,9 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-android.enableBuildCache=true
 android.enableJetifier=true
 android.useAndroidX=true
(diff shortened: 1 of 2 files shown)


Mygod:
According to Travis, commit d637392 [took 17:46](https://travis-ci.org/shadowsocks/shadowsocks-android/builds/408855809?utm_source=github_status&utm_medium=notification) and this [took 18:27](https://travis-ci.org/shadowsocks/shadowsocks-android/builds/408950311?utm_source=github_status&utm_medium=notification). Maybe the extra cache thing slowed it down? Is it possible to include those cache in travis file as well?

pexcn:
Yes, I add this line:
```
        - $HOME/.android/build-cache/
```

reference: https://docs.travis-ci.com/user/languages/android/#caching

pexcn:
According to travis-ci docs:
https://docs.travis-ci.com/user/languages/android/#gradle-daemon-is-disabled-by-default

but gradle docs was updated, now run daemon by default:
https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:disabling_the_daemon

So we should run gradle as daemon and enable parallel build?
```properties
org.gradle.daemon=true
org.gradle.parallel=true
```

Mygod:
IIRC Travis doc is pretty outdated. You should probably double check if that still applies.

gradle.properties
```diff
@@ -9,11 +9,12 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
+android.enableBuildCache=true
```
aaronhe42: this is already default to `true` in AGP 2.3+

pexcn:
I just updated.

Mygod:
Still didn't see any improvements in Travis build after restarting the build. But the code looks fine.

pexcn:
maybe need at least twice build?

PR #1913, opened 2018-08-15
Remove night mode switch

Mygod:
It seems that Android 9.0 has finally supported Night mode officially. Unfortunately it doesn't work on emulator (not on my side) so I'm not sure if this works.

@onlymash Please review.

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/App.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/App.kt
@@ -136,7 +136,7 @@ class App : Application() {
                 }
             DataStore.publicStore.putLong(Key.assetUpdateTime, info.lastUpdateTime)
         }
-        AppCompatDelegate.setDefaultNightMode(DataStore.nightMode)
+        if (Build.VERSION.SDK_INT < 28) AppCompatDelegate.setDefaultNightMode(DataStore.nightMode)
         updateNotificationChannels()
     }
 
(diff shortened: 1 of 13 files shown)


onlymash:
9.0 的夜间模式设置在开发者选项里，原来的方案一切正常，不用改。
screenshot_20180825-130021

onlymash:
不对，如果把系统的夜间模式设为自动，在 App 内的切换，自动和跟随系统便无法获取当前系统主题状态了

onlymash:
在 8.1 系统， 通过 UiModeManager 获取的 nightMode 要么是 1 （关闭）要么是 2 （开启）。9.0 系统开发者选项里把系统夜间模式设置为自动后 UiModeManager 获取的 nightMode 值是 0，于是 App 无法判断当前系统夜间模式是开启还是关闭，导致修改 App 的自动和跟随系统，当前的 App 主题不能立即生效

Mygod:
Looking at `AppCompatDelegateImpl.java:2099`:

```
config.uiMode = newNightMode | (config.uiMode & ~Configuration.UI_MODE_NIGHT_MASK);
```

The old UI mode from the system is wiped when new night mode is applied. It just seems to me that support library just doesn't consider changing night mode at runtime a possibility.

Also setting night mode to `MODE_NIGHT_AUTO` would technically require LOCATION permissions to function properly (see `TwilightManager`) unless system night mode was also set to auto since API 23 (instead of 28).

I don't want too much code around just to handle support library not being able to handle this case and permission issues so I think it's better to just remove the switch completely.

onlymash:
只保留开启和关闭两个选项，大多数应用都是这么做的

PR #1948, opened 2018-09-09
Allow show password

Mygod:


Mygod:
DO NOT MERGE. Succeeded by https://github.com/Gericop/Android-Support-Preference-V7-Fix/pull/182.

Mygod:
@madeye Do you think this feature would be a good addition at all?

madeye:
Maybe not, as the password would not be readable at all.

Mygod:
Okay.

Mygod:
@madeye Hmm, why would the password be not readable?

Mygod:
@madeye I'm considering reviving this PR. Currently the profiles can already be exported so I don't see why we need to hide the password.

madeye:
I think we need to hide password by default, as a best practice. 

Showing password would be useful, but optional.

PR #1982, opened 2018-10-26
don't set GOROOT if already exist

xmh19936688:


xmh19936688:
`%DIR%\go` don't exist after clone. And I have go env on my PC. So I try like this and `BUILD SUCCESSFUL in 1m 29s`

madeye:
Custom GO is not needed any more. Please update this PR to remove custom Go part.

xmh19936688:
Like this?

PR #1990, opened 2018-11-03
Fix bottom app bar

Mygod:
TODO: This version introduces [another crash on Android 9.0](https://issuetracker.google.com/issues/118921801).

Also fixes [toolbar layout on tablets](https://issuetracker.google.com/issues/112714499).


madeye:
Hmm, let's wait for an alpha2 release...

Mygod:
Ummm I don't think alpha2 is released yet?

madeye:
Hmm, let me revert the merge first. It looks not a easy task to do a hacking here.

PR #1996, opened 2018-11-09
make code clear

xmh19936688:


madeye:
It's type is `Boolean?`.

PR #2020, opened 2018-12-07
Fix bottom app bar

madeye:
With a snapshot AAR from self hosted maven repo.


madeye:
Tested on Android pie. Now everything works well.

PR #2048, opened 2018-12-27
Replace DNS forwarding with UDP forwarding

madeye:
This PR aims to let user decide if forward UDP traffic to remote server.

### DIFF
--- a/core/src/main/res/values-zh-rCN/strings.xml
+++ b/core/src/main/res/values-zh-rCN/strings.xml
@@ -31,8 +31,8 @@
 <string name="auto_connect_summary">"允许 Shadowsocks 随系统启动"</string>
 <string name="tcp_fastopen_summary">"切换可能需要 ROOT 权限"</string>
 <string name="tcp_fastopen_summary_unsupported">"不支持的内核版本: %s &lt; 3.7.1"</string>
-<string name="udp_dns">"DNS 转发"</string>
-<string name="udp_dns_summary">"转发所有 DNS 请求到远程服务器"</string>
+<string name="udp_dns">"UDP 转发"</string>
+<string name="udp_dns_summary">"转发所有 UDP 请求到远程服务器"</string>
 
 <!-- notification category -->
 <string name="forward_success">"后台服务已开始运行。"</string>
--- a/core/src/main/res/values/strings.xml
+++ b/core/src/main/res/values/strings.xml
@@ -57,8 +57,8 @@
   <string name="direct_boot_aware_summary">Your selected profile information will be less protected</string>
   <string name="tcp_fastopen_summary">Toggling might require ROOT permission</string>
   <string name="tcp_fastopen_summary_unsupported">Unsupported kernel version: %s &lt; 3.7.1</string>
-  <string name="udp_dns">DNS Forwarding</string>
-  <string name="udp_dns_summary">Forward all DNS requests to remote</string>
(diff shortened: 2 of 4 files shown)


madeye:
I think it's useful for VoIP and mobile gaming.

Mygod:
I think this is a bad idea. Filtering traffic by transport layer protocol is likely to break applications that use both.

Oneeyear:
hey，bro，what is wrong with you，please do not send email to me again，thank you！





            
            
                
                    
                            
                                
                            
                            
                                pchmx
                            
                    
                        
                            
                                    邮箱：pchmx@qq.com
                            
                        
                
            
        Signature is customized by Netease Mail Master


On 12/27/2018 14:38, Mygod wrote: I think this is a bad idea. Filtering traffic by transport layer protocol is likely to break applications that use both.

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/shadowsocks/shadowsocks-android","title":"shadowsocks/shadowsocks-android","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/shadowsocks/shadowsocks-android"}},"updates":{"snippets":[{"icon":"PERSON","message":"@Mygod in #2048: I think this is a bad idea. Filtering traffic by transport layer protocol is likely to break applications that use both."}],"action":{"name":"View Pull Request","url":"https://github.com/shadowsocks/shadowsocks-android/pull/2048#issuecomment-450086684"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/shadowsocks/shadowsocks-android/pull/2048#issuecomment-450086684",
"url": "https://github.com/shadowsocks/shadowsocks-android/pull/2048#issuecomment-450086684",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

madeye:
IMO, different UDP/TCP source address won't be a big problem, especially for online gaming. Also, it'd be helpful if someone are using a TCP only shadowsocks server.

Let's keep it open for now, until we figure out if it's really useful.

Mygod:
I still think per-app VPN is a better solution. I don't see how this is useful in any case.

Mygod:
Here's another reason why this is a bad idea: this makes it extremely easy to make a seemingly innocent app that enables the developer to determine your proxy's IP address. As no other apps use such mechanism AFAIC, this also helps the server to determine the type of the proxy used.

sqwwqw5:
It's not a good idea to simply replace DNS Forwarding to UDP Forwarding from my point of view. Why not having both? In that case,just leave the UDP53 forwarded when user choose to disable UDP Forwarding or replace UDP to TCP for DNS Forwarding.

madeye:
Let's close it for now.

Mygod:
Since some version (I think 3.x), UDP forwarding is always enabled. This PR can be better described as UDP bypassing.

PR #2066, opened 2019-01-08
fix Windows building bug

CGDF-Github:

### DIFF
--- a/core/src/overture/make.bat
+++ b/core/src/overture/make.bat
@@ -28,7 +28,7 @@ MKDIR %TARGET%\armeabi-v7a>nul 2>nul
 MKDIR %TARGET%\x86>nul 2>nul 
 MKDIR %TARGET%\arm64-v8a>nul 2>nul 
 
-SET CC=%ANDROID_ARM_TOOLCHAIN%\bin\arm-linux-androideabi-gcc.exe
+SET CC=%ANDROID_ARM_TOOLCHAIN%\bin\arm-linux-androideabi-gcc.cmd
 
 IF NOT EXIST %ANDROID_ARM_CC% (
 	ECHO "Make standalone toolchain for ARM arch"


Mygod:
This looks good. Can you try NDK r19 and [update the build script](https://github.com/shadowsocks/shadowsocks-android/commit/e005c81eeec67cd554eb6c1da5fd588eec3fbf38) while you're at it? 😄

CGDF-Github:
> This looks good. Can you try NDK r19 and [update the build script](https://github.com/shadowsocks/shadowsocks-android/commit/e005c81eeec67cd554eb6c1da5fd588eec3fbf38) while you're at it? 😄

I'm in r19.
In addition, AS3.3 will install Gradle 5.0 for this project. People must downgrade to 4.x to build on Windows.

CGDF-Github:
I'm sorry. I wrote some mistakes in my last comment. I have edited it.
I built with lastest commit(https://github.com/shadowsocks/shadowsocks-android/commit/9099e58e8ac8075949897aed62cad3bf3af7afaf) on Windows and succeeded.

One more thing, i'm not sure whether it will be better to remove this line in `clean.bat`:


`clean.bash` does not do this. And this line will give a warning (cannot find this folder).
I think it is a historical legacy.

Mygod:
Yes. `clean.bash` and `make.bash` has been updated for NDK r19. I believe you shouldn't be able to build with old `clean.bat` and `make.bat` once you do a clean since we are no longer using standalone toolchains in NDK r19. It would be great if you can update those `bat` scripts. 😆

CGDF-Github:
You are right, the old `make.bat` doesn't work. I have spent hours in trying to update the build script on Windows since yesterday. But now I'm stuck in a error and i have no idea how to solve it.

This pr is meaningless. I misunderstood something.

CGDF-Github:
I wasted several days on it.
https://github.com/golang/go/issues/17014
I thought this issue of golang may be solved because it has been two years.
I was wrong...

Golang's engineers are pigeons...

Mygod:
Hmmm, I thought the old `make.bat` also uses clang?

CGDF-Github:
> Hmmm, I thought the old `make.bat` also uses clang?

i don't think old `make.bat` worked correctly in the past.
nobody submitted issue about `make.bat` just because it seems can be built correctly.
`go build` error returns 0.... so, people will not get any notice when they make project.
and people can build apk without any error.
the apk can be installed and launched, but it cannot start VPN service.
little people will find the reason is `make.bat`.

that's my guess, i need sometime to confirm.

the old `make.bat` looks really strange.
```
SET CC=%ANDROID_ARM_TOOLCHAIN%\bin\arm-linux-androideabi-gcc.exe`
IF NOT EXIST %CC% (
    ECHO "gcc not found"
    EXIT 1
)
```
but it never use this value...

Mygod:
I agree. I guess building on Windows does not work after all. :sweat_smile:

Mygod:
Could you try building with Docker though? Maybe something like: `docker run --rm -v ${PWD}:/build shadowsocks/shadowsocks-android:circleci bash -c "cd /build; git clone https://github.com/shadowsocks/shadowsocks-android; cd shadowsocks-android; git submodule update --init --recursive; ./gradlew goBuild"`

CGDF-Github:
You mean... on Windows?

Mygod:
Yes. I believe Windows has docker.

CGDF-Github:
I think running `goBuild` in Docker will be successfully. I will try it.

PR #2072, opened 2019-01-14
Plugin library 1.1.0

Mygod:
Okay. Our old `beta`/`develop` branches definitely went nowhere. Let's just replace `beta` with feature PRs and `develop` with PRs like this one. In theory we should release an update for plugin lib immediately after merging PRs like this one.


madeye:
Updated on sonatype.

Mygod:
That was fast... I was originally thinking doing a new release accompanying a new release of the main app but I guess this works too. :sweat_smile:

Issue #2083, opened 2019-01-19
启动后经常会卡死几次后，才能正常使用


xcjiang:
Please read contributing guidelines. Thanks.

**Describe the bug**
不确定从那个版本开始，有一个奇怪的现象：每次重新启动ss后，总会遇到好几次卡死。
具体是，启动SS后，正常使用一段时间（几分钟或十几分钟）， 然后国外APP就不通了，这时你就知道有问题了。此时进入SS界面，点击下方的绿色图标来关闭当前连接，发现无任何反应，关不掉。
这时候任何办法都无法恢复，除非进入SS应用的信息页，点击强制停止后，再重新启动SS。以上会重复4，5次后，然后会进入稳定状态，此时用上十几天也不会出任何问题。然后到下一次关闭重启SS后，就会要再折腾这么一番才能正常使用。

**To Reproduce**
1. 启动SS
2. 点击下方圆图标启动
3. 正常使用手机一段时间, 此时SS工作正常，可正常翻墙
4. 几分钟或十几分钟后，SS忽然停止工作，不能翻墙
5. 进入SS界面
6. 点击绿色图标
7. 无任何反应，无法关闭，感觉SS已经被卡死。

**Expected behavior**
不会出现上述的问题

**Screenshots**

**Smartphone (please complete the following information):**
 - Android/version: 8.0
 - Device: Samsung S8
 - Version: 4.7
 - Last version that did not exhibit the issue: [not applicable]

**Configuration**

* [ x] IPv4 server address
* [ ] IPv6 server address
* [ x] Client IPv4 availability
* [ ] Client IPv6 availability
* Encrypt method:
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [x ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [ x] Apps VPN mode
  * [ x] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS Forwarding
* Plugin configuration (if applicable):
* [ ] Auto Connect
* [ ] TCP Fast Open
* If you're not using VPN mode, please supply more details here:

**Additional context**
如需要其他信息，请告知

AlvinSeyir:
Same problem here. I use Samsung note 8 with android 8 system.

Mygod:
[...]

PR #2089, opened 2019-01-24
Refactor using coroutines

Mygod:
What are coroutines: Basically threads but much more light-weight. `Dispatchers.Main` is implemented using Android's `Handler/Looper` mechanism.

With this PR, most bg work can be done on the main thread, so we have fewer concurrency issues to worry about. Furthermore, sockets are now protected in parallel using IO thread pool and error callbacks are refined. (see commit logs for more info)

This PR also cleans up various IO threads that live longer than that they are supposed to live. Fewer threads = less RAM!

One downside of using coroutines, however, is that debugging deadlocks can be much harder, as dumping stacks aren't useful. This might be mitigated using [kotlinx-coroutines-debug](https://github.com/Kotlin/kotlinx.coroutines/tree/master/core/kotlinx-coroutines-debug) but that needs more investigation.

This PR might help with #2083.

I don't think this will impact binary size that much as `android.arch.work:work-runtime-ktx` already uses Kotlin coroutines.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,6 @@ apply plugin: 'com.github.ben-manes.versions'
 buildscript {
     ext {
         kotlinVersion = '1.3.20'
-        kotlinCoroutinesVersion = '1.1.0'
         minSdkVersion = 21
         sdkVersion = 28
         compileSdkVersion = 28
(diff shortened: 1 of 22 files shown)


madeye:
LGTM

PR #2092, opened 2019-01-27
Use dnsproxyd for DNS resolving

### LINKED ISSUE #2016: about local-dns global config
We have had a global config named 'Local DNS port'. So, the new config name will be 'Customize Local DNS upstream'? If user enable it, then he/she need input a string such as 'udp:114.114.114.114,udp:119.29.29.29:53,tcp-tls:cloudflare-dns.com,tcp-tls:dnsovertls.sinodun.com:443@145.100.185.15'??

Mygod:
:construction: DO NOT MERGE :construction: 

This fixes occasional DNS resolving timeouts when switching profiles in VPN mode, due to race conditions where the DNS request was sent to the VPN tun interface shortly before its tear-down. This also enables the possibility that we can switch profile without recreating another tun interface (contributions are welcome).

However, in order to make this work, we need to know default network other than current VPN interface. Therefore, we will encounter [`WRITE_SETTINGS` permission bug](https://stackoverflow.com/a/33509180/2245107) on Android 6.0(.0). The current workaround is to use `registerNetworkCallback` so that we will use any default network candidate (possibly mobile data while Wi-Fi is present) as our default network. We also show a Toast prompting user to enable the permission (further improvements are welcome).

We can also respect local DNS settings (fix #2016), and therefore including possibly support private DNS on Android 9+, by using dnsproxyd to do all the lookup for us which I am working on currently. Meanwhile, @madeye What do you think?

P.S. It seems like private DNS is not working on my side with Wi-Fi + VPN but works fine with mobile data + VPN. Can anyone also confirm this?

### DIFF
--- a/core/src/main/AndroidManifest.xml
+++ b/core/src/main/AndroidManifest.xml
@@ -9,6 +9,8 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
+    <!-- This permission is only used on Android 6.0 due to bug: https://stackoverflow.com/a/33509180/2245107 -->
+    <uses-permission-sdk-23 android:name="android.permission.WRITE_SETTINGS"/>
 
     <uses-feature android:name="android.software.leanback"
                   android:required="false"/>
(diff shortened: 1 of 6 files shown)


madeye:
So you're planning setup a local DNS resolver querying from `netd`?

Mygod:
Yes, using the infrastructure in this PR, basically replacing overture entirely.

This has already proven to be a huge headache (probably another 300-400 lines of code thanks to dnsjava?) and I am already regretting this just like I regretted #1478. :cry:

PR #2096, opened 2019-01-29
New local DNS server implemented with Kotlin coroutines

Mygod:
~:construction: DO NOT MERGE :construction:~ ~No longer in construction but in active testing~

Supersedes https://github.com/Mygod/shadowsocks-android/pull/2.

> ~Nothing works yet (well actually receiving DNS and local resolver work) and somebody just needs to debug this I guess.~

> Also I believe Kotlin coroutines have comparable performance with goroutines [[1]](https://medium.com/@omkard/kotlin-ktor-vs-golang-performance-eb3c56127373) [[2]](https://stackoverflow.com/questions/46864623/which-of-coroutines-goroutines-and-kotlin-coroutines-are-faster) [[3]](https://proandroiddev.com/kotlin-coroutines-vs-rxjava-an-initial-performance-test-68160cfc6723), but obviously that needs fine tuning. The current code probably runs a little slow but again code in overture was not written with great care either.

> [With this PR, we] can also respect local DNS settings (fix https://github.com/shadowsocks/shadowsocks-android/issues/2016), and therefore including possibly support private DNS on Android 9+ (closes https://github.com/shadowsocks/shadowsocks-android/issues/2015).
>
> ~P.S. It seems like private DNS is not working on my side with Wi-Fi + VPN but works fine with mobile data + VPN. Can anyone also confirm this?~

Update: https://github.com/shadowsocks/shadowsocks-android/pull/2096#issuecomment-460022145


Mygod:
@madeye Do you think this would be a good idea at all (assuming all bugs fixed)? There is also an alternative approach, which is to pass network handle of `underlyingNetwork` to overture and overture will call [`android_getaddrinfofornet`](https://developer.android.com/ndk/reference/group/networking#group___networking_1gadbe9af25952513a61f3ee72dd1395570). That way I believe there is also no need to use protect socket at all for overture. (but I really don't like including a heavy golang project fork if it can be done with ~300 lines of Kotlin)

madeye:
I think it depends on the effort we need to implement a kotlin local DNS server.

If we can solve the problem with existing software and much less effort, it'd be better try it first.

Mygod:
Everything fixed (I believe). Still <200 lines. :laughing:

Mygod:
Now the universal apk is down to ~6MB (split apk ~5MB), also [CircleCI builds](https://circleci.com/gh/Mygod/shadowsocks-android/98).

madeye:
Cool! I will do more tests later.

Mygod:
There are still a few glitches with `withTimeout` it seems. Meanwhile, I am trying to rewrite this (yet again) using non-blocking IO (`java.nio.channels`) to see if we can get around that.

I should probably stop changing stuff in this repo every two days and focus on other work... Please send help.

madeye:
No hurry, it's not an urgent task. 🤗

madeye:
Cool! BTW, I didn't have glitches locally. Maybe just decrease the timeout here.

Mygod:
Well, the glitch is that sometimes service will be stuck in STOPPING state. I only encountered it once though and I'm not sure if decreasing timeout would help.

madeye:
I'm testing on a Huawei device.

A continuous DNS resolver failure happens after one day use (running in background for more than 24 hours). Not sure if it's because background processes killed by the system.

Will try to reproduce the issue today.

madeye:
I just found plugin drop-down menu cannot be opened now. Anything changed to plugin logic recently?

Mygod:
What dropdown? Everything seems fine here.

Mygod:
:rofl: I would suggest a clean build.

madeye:
It looks related to the R8 shrinker. After setting `android.enableR8.fullMode` to `false`, I can get the issue fixed.

How to check if my local R8 is up to date?

Mygod:
Okay, that makes sense because we used `PreferenceFragmentCompat.registerPreferenceFragment`.

I think the difference should be ~50KB. I will try to remove the problematic takisoft library soon.

madeye:
It looks if the phone goes into deep sleep mode, the local DNS server cannot be waked up any more.

The steps to reproduce

1. Put your phone on a desk.
2. Wait for 7 hours.
3. Unlock your phone.
4. Visit 1.1.1.1, and it works.
5. Visit google.com, not working.

Mygod:
> 2. Wait for 7 hours.

Yeah right.

Mygod:
Just tested it with dumpsys doze. No issues observed. :confused:

madeye:
With 0efe2e8, stuck on connecting is observed. 

With 00dfb92, stuck on stopping is observed again.

madeye:
Steps to reproduce the problem of https://github.com/shadowsocks/shadowsocks-android/commit/0efe2e8aaf3d3f5f5b48ef4aabb76bf3659d9757,

1. Connect to a server.
2. Disconnect.
3. Connect again.
4. Get stuck.

Mygod:
How do you reproduce https://github.com/shadowsocks/shadowsocks-android/commit/00dfb925c71ed73ec687d9e8c0fdb97f5a83cdaa?

madeye:
Okay, I think 00dfb92 brought a lot of issues here...

madeye:
With 00dfb92,

1. Connect to a server.
2. Browser some websites for a while.
3. Lock the phone for a while.
4. Unlock the phone.
5. All the websites cannot be resolved.
6. Try disconnecting.
7. Get stuck.

madeye:
BTW, all the tests are done with v2ray plugin.

madeye:
The key to reproducing issue of 00dfb92 is locking the phone. So, it looks related to doze?

madeye:
Interesting, looks 493b6e8 fixed all the issues.

Let me do more tests locally.

Mygod:
It's probably helpful to dump the thread stacks for debugging when that happens: https://stackoverflow.com/q/13589074/2245107

Mygod:
BTW https://github.com/shadowsocks/shadowsocks-android/commit/493b6e856de680e0ce9b5989b73fcc680276d8b0 actually fixes bugs introduced by https://github.com/shadowsocks/shadowsocks-android/pull/2096/commits/490e04323e221090109145ccb80288d87b72ea06.

Mygod:
@madeye https://github.com/shadowsocks/shadowsocks-android/pull/2096/commits/f4e65b2eedebfd9f58cee58c3ad264eb239063d2 and https://github.com/shadowsocks/shadowsocks-android/pull/2096/commits/9894d0e9a5cfcf6d4bb70e0ae1d217b4172cc64a should fix occasional deadlocks on connecting.

There are quite a few commits in the last few hours that should have been committed to master branch instead, but I am too lazy to switch branches and do merges. :sweat_smile:

Mygod:
@madeye Any updates? It is running super smoothly for me for now.

madeye:
Yeah, i think all the issues have been resolved, including previous doze problems.

madeye:
Yeah, I think so.

Mygod:
No complaints heard over the past 3 days. I am pretending that this is good enough because YOLO! :rofl:

madeye:
It looks the deep sleep problem happened again.

With the latest commit, the DNS resolver died after hours of standby.

PR #2160, opened 2019-03-15
Support metered hint on Android 9+

Mygod:
This is an [Android Q new feature](https://developer.android.com/reference/android/net/VpnService.Builder.html#setMetered(boolean)) but it can be backported to Android 9 using a hack. Adaptation for Android Q will be added to `q-beta` branch.


18813961612:
it is ok

PR #2193, opened 2019-04-23
Add a key to settings screen

onlymash:
Set a key on the PreferenceScreen if using initialExpandedChildrenCount. 
the state is correctly saved and restored when the configuration changes (such as when rotating the screen).


Mygod:
Looks good. Thanks.

Issue #2182, opened 2019-03-30
夜间主题下高级设置选项的图标颜色问题


onlymash:
RT
![IMG_20190330_173836](https://user-images.githubusercontent.com/5838634/55274411-ed769880-5312-11e9-966d-15ace6f32bdd.jpg)

Mygod:
This is an upstream bug. Contributions are welcome.

williamxxy:
？？？？这个是因为没有root权限，只需要root就行

Mygod:
This is fixed in preference lib 1.1.0-alpha05.

PR #2194, opened 2019-04-23
Fix tint color #2182

onlymash:


core/.gitignore
```diff
@@ -1 +1,2 @@
 /build
+.cxx/
```
Mygod: What's this for?

onlymash: I using AS 3.5, automatically generate this folder.

onlymash:
Advanced is an ExpandButton, and ExpandButton doesn't provide any public methods, so we can only modify the global tint

onlymash:
A hide class [ExpandButton](https://android.googlesource.com/platform/frameworks/support/+/c2923b8daa8f500075fad0586c9901d44b49ded5/preference/src/main/java/androidx/preference/ExpandButton.java)

Mygod:
Can we try replacing ic_arrow_down_24dp?

core/.gitignore
Mygod: What are the steps to reproduce this issue?

onlymash: AS 3.5 NDK 20.0.5392854-beta2
[.cxx dir](https://i.loli.net/2019/04/25/5cc1d2e734918.png)

Mygod: Okay. I'll check again when NDK r20 goes stable.

plugin/build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     api 'androidx.core:core-ktx:1.0.1'
     api 'com.google.android.material:material:1.1.0-alpha05'
     api "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    api "com.takisoft.preferencex:preferencex-simplemenu:$preferencexVersion"
```
Mygod: Why?

onlymash: PreferenceThemeOverlay depend preference

Mygod: In that case, moving the fix to core is preferred.

PR #2196, opened 2019-04-26
Refine auto connecting

Mygod:
Here's some further possible enhancements for this feature:

* Auto connect only if the service was not shut down manually by user, i.e. via stop broadcast or other VPN app replacing, but not package upgrades/force stops/system shutdowns.
* If we did this, maybe we can remove auto connect switch altogether?

@madeye Thoughts?


madeye:
I think `auto connect` is a legacy option, as we didn't have always-on VPN and tasker integration years ago.

So, I'm find to remove it in the future.

Mygod:
Here's my experience with always-on VPN and the motivations for my proposed change:

1. You need to go to the settings to switch to another VPN app. (I need to do this occasionally as I need to test [VPN Hotspot](https://github.com/Mygod/VPNHotspot) with a variety of different VPN apps)
2. You also need to go to the settings to turn off always-on VPN if you are travelling and you no longer need to use VPN.
3. Always-on VPN tends to confuse user as it forbids other VPN apps from establishing a VPN connection altogether (`VpnService.establish` simply returns null) and the error message is often unclear.

madeye:
Yeah, always-on VPN is a little bit confusing, and very few users understand how it works.

But if we remove the "auto connect" option and enable this by default, I think it may also introduce some confusion, especially if we cannot have different scenarios well.

If we just "auto connect" on boot, maybe we can simplify the logic and remove the option safely.

Mygod:
What? How come it has already been 25 days? 😅

Mygod:
@madeye Ready for (final) review. Please see if this new behavior makes sense.

PR #2198, opened 2019-04-28
ipv6 use reserved addr defined in rfc4291

boypt:
# Problem

`fdfe:dcba:9876::1` is an ULA, defined in RFC 4193, which meant for communications among private IPv6 networks.

But in the scenario of VPN application, the interface addresss is actually for NAT6, private address would be translated to communicate with remote servers. This isn't what RFC 4193 ULA meant for.

As result, even when properly configured, open https://test-ipv6.com in browser, still getting a warnning:

```
Your browser has real working IPv6 address - but is avoiding using it. We're concerned about this. [more info]
```

That's caused by apps faceing dual-stack servers, an ULA address had a very low priolity and apps tend to avoid it.

# Resolving

Simply change the ULA address to a reserved block, eg `c000::/3`. ref:  [IANA Address Space](https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml)

# More info

 * [论 NAT6（IPv6 NAT）](https://zhuanlan.zhihu.com/p/38680594)
 * [Openwrt User guide for NAT6](https://openwrt.org/docs/guide-user/network/ipv6/ipv6.nat6#faq_and_troubleshooting) #Why should the ULA prefix be changed?

### DIFF
--- a/core/src/main/java/com/github/shadowsocks/bg/VpnService.kt
+++ b/core/src/main/java/com/github/shadowsocks/bg/VpnService.kt
@@ -56,8 +56,8 @@ class VpnService : BaseVpnService(), LocalDnsService.Interface {
         private const val VPN_MTU = 1500
         private const val PRIVATE_VLAN4_CLIENT = "172.19.0.1"
         private const val PRIVATE_VLAN4_ROUTER = "172.19.0.2"
-        private const val PRIVATE_VLAN6_CLIENT = "fdfe:dcba:9876::1"
-        private const val PRIVATE_VLAN6_ROUTER = "fdfe:dcba:9876::2"
+        private const val PRIVATE_VLAN6_CLIENT = "ddfe:dcba:9876::1"
+        private const val PRIVATE_VLAN6_ROUTER = "ddfe:dcba:9876::2"
 
         /**
          * https://android.googlesource.com/platform/prebuilts/runtime/+/94fec32/appcompat/hiddenapi-light-greylist.txt#9466


Mygod:
Does your fix resolve the problem you mentioned?

boypt:
Changing the address to `ddfe:dcba:9876::1` resolves the problem. `ddfe:dcba:9876::1` belongs to `c000::/3`.

test-ipv6 wont show the warnning info.

Mygod:
> #### Why should the ULA prefix be changed?
>
>The default ULA prefix starting with represents an address that is not globally routed on the internet. Some (or most) clients will prefer the IPv4 route, if they don't have a global IPv6 address assigned, so you need to change the prefix to indicate a global address. It doesn't necessarily have to start with , but to avoid conflicts, you should use a prefix that is not being used yet. The letters are unassigned and therefore safe choices.

I think using ULA prefix makes sense here and it's a good idea to make clients aware instead of tricking them.

Mygod:
I reconsidered this again and I think what we currently have is the best approach for shadowsocks. If you want to bypass low priority issue, you should use your patch (though I would not recommend it). If you want a globally routable IPv6 address, you should probably use WireGuard.

On that note, obfuscating WireGuard traffic sounds like an interesting problem. A possible solution is to transmit WireGuard over QUIC. @madeye Thoughts? :rofl:

madeye:
Wireguard is kind of whitelisted by GFW now.

Given wireguard will soon become a standard of cloud infrastructure, I doubt anyone would block that protocol. Similarly, we never see Cisco anyconnect is blocked.

About the UDP obfuscation, we have a lot of choices, a fake QUIC obfuscation should be doable.

PR #2211, opened 2019-05-18
Fix #2209

### LINKED ISSUE #2209: 手动设置配置信息时 换屏信息将不会保留
手动设置配置信息时 会一下子记不住配置信息，需要返回只别的界面复制剩余信息，但重新返回APP时 原先设置了一半的界面消失了，app自动返回至首页。
原来记得3.xx的版本 返回APP时仍是原先的界面，但升级至4.7.4之后，就有这个问题了

Mygod:
This is a basic fix for #2209.

This has an unwanted side effect: `MainActivity` will no longer respond to `android.intent.action.VIEW` (for `ss` URIs) if it is not in the top of the task. This intent should probably be handled in a separate `Activity` or sort.

Relevant links for why such fix:

* https://developer.android.com/guide/topics/manifest/activity-element#lmode
* https://stackoverflow.com/a/2059480/2245107
* https://stackoverflow.com/a/30821062/2245107

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ apply plugin: 'com.github.ben-manes.versions'
 
 buildscript {
     ext {
+        javaVersion = JavaVersion.VERSION_1_8
         kotlinVersion = '1.3.31'
         minSdkVersion = 21
         sdkVersion = 28
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -28,6 +28,11 @@ android {
         }
     }
 
+    compileOptions {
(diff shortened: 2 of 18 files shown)


madeye:
Looks a valid fix.

Is it possible we still use singleTask mode but restore fragment states when the activity is started again by the launcher?

Mygod:
Do you mean restore task back stack? I don't know but I imagine that would be more work.

madeye:
Yeah, I think so. Let's go this solution.

PR #2221, opened 2019-05-29
fix https://github.com/shadowsocks/shadowsocks-android/issues/2215

jhoneybee:
I added a small feature to test the node's network  fix https://github.com/shadowsocks/shadowsocks-android/issues/2215

image

### DIFF
--- a/core/src/androidTest/java/com/github/shadowsocks/utils/NetworkUtilsTest.kt
+++ b/core/src/androidTest/java/com/github/shadowsocks/utils/NetworkUtilsTest.kt
@@ -1,0 +1,13 @@
+package com.github.shadowsocks.utils
+
+import org.junit.Assert
+import org.junit.Test
+
+
+class NetworkUtilsTest {
+    @Test
+    fun pingTest(){
+       val result =  NetworkUtils.ping("127.0.0.1")
+        Assert.assertEquals(result.type,true)
+    }
+}
--- a/core/src/main/java/com/github/shadowsocks/utils/NetworkUtils.kt
+++ b/core/src/main/java/com/github/shadowsocks/utils/NetworkUtils.kt
@@ -1,0 +1,75 @@
+package com.github.shadowsocks.utils
+
+import android.R.attr.host
+import java.io.BufferedReader
+import java.io.InputStreamReader
+import java.util.regex.Pattern
+
+
+/*******************************************************************************
+ *                                                                             *
+ *  Copyright (C) 2018 by https://github.com/shadowsocks/shadowsocks-android   *
+ *  Copyright (C) 2018 by https://github.com/shadowsocks/shadowsocks-android   *
+ *                                                                             *
+ *  This program is free software: you can redistribute it and/or modify       *
+ *  it under the terms of the GNU General Public License as published by       *
+ *  the Free Software Foundation, either version 3 of the License, or          *
+ *  (at your option) any later version.                                        *
+ *                                                                             *
+ *  This program is distributed in the hope that it will be useful,            *
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of             *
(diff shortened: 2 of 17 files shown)


madeye:
You should no rely on ping to do the test.

Actually, the test should take advantage of ss-local to perform a real connection test, e.g. connect to `google.com` through the proxy.

jhoneybee:
> You should no rely on ping to do the test.
> 
> Actually, the test can take advantage of `ss-local` to perform a real connection test, e.g. connect to `google.com` through the proxy.

 I'm not very familiar with ss-local. I'll try to use this.

> can I use ```SonarLint``` plug-in as a code specification?

jhoneybee:
@madeye  
I found that my code specification did not conform to ```Codacy/PR Quality Review'.``` 

> Is there a plug-in for Android studio?

Mygod:
That is a very wasteful way to do ping as well. If you just want to do a ping test, you can easily write a bash script that is much easier.

jhoneybee:
> That is a very wasteful way to do ping as well. If you just want to do a ping test, you can easily write a bash script that is much easier.

Thank you very much for correcting my mistake.


Then I will do two things 

1. use ```ss-local``` test connection 
2. test network latency in other ways

Is there any better advice for testing network latency?

jhoneybee:
> That is a very wasteful way to do ping as well. If you just want to do a ping test, you can easily write a bash script that is much easier.

@Mygod  Excuse me, an additional topic, what plug-ins do you use to check code specifications?

Mygod:
If you would still like to implement this feature, I would suggest to focus on implementing it first.

And to answer your question directly, no.

Mygod:
To be honest, I don't see why you care about batch testing latency (whose expected value hardly ever changes) instead of throughput, jitter, etc.

jhoneybee:
> To be honest, I don't see why you care about batch testing latency (whose expected value hardly ever changes) instead of throughput, jitter, etc.

Thank you very much. I'm an application-level developer. I don't know much about the underlying code, or even the network jitter. I'll find out about this convenient information.


> I'll close this request first and implement it in a better way in the future. Now I'm going to learn about ```ss-local``` code.

jhoneybee:
> You should no rely on ping to do the test.
> 
> Actually, the test can take advantage of `ss-local` to perform a real connection test, e.g. connect to `google.com` through the proxy.

@Mygod  I'm sorry, but I'd like to ask you how to call dynamic link libraries in the project.

>  Sorry, I am a newbie. It doesn't seem to be using JNI pattern calls

jhoneybee:
@Mygod I'm sorry, but I'd like to ask you how to call dynamic link libraries in the project.

> Sorry, I am a newbie. It doesn't seem to be using JNI pattern calls

diov:
> @Mygod I'm sorry, but I'd like to ask you how to call dynamic link libraries in the project.
> 
> > Sorry, I am a newbie. It doesn't seem to be using JNI pattern calls

Actually, they build the executable binaries but name them with *.so suffix. It's tricky but very useful, LOL.
So you can call the dynamic link libraries with `ProcessBuild` or `Runtime.exex()` easily. Also, you can check the [Makefile script](https://github.com/shadowsocks/shadowsocks-android/blob/master/core/src/main/jni/build-shared-executable.mk) to find how it works.

jhoneybee:
> > @Mygod I'm sorry, but I'd like to ask you how to call dynamic link libraries in the project.
> > > Sorry, I am a newbie. It doesn't seem to be using JNI pattern calls
> 
> Actually, they build the executable binaries but with *.so suffix naming. It's tricky but very useful, LOL.
> You can check the [Makefile script](https://github.com/shadowsocks/shadowsocks-android/blob/master/core/src/main/jni/build-shared-executable.mk) to find how it works.
> So you can call the dynamic link libraries with `ProcessBuild` or `Runtime.exex()` easily, without JNI pattern calls.

I have been looking for jni mode calls, the direction is wrong, thank you very much

Issue #2234, opened 2019-06-08
v4.8.0 crash on Apps VPN mode


m5h1:
**Describe the bug**
The application crashes when trying to save Apps VPN Mode settings 

**To Reproduce**
1. Open the application 
2. Add Profile --> Manual Settings --> Apps VPN Mode 
3. Turn on Apps VPN Mode for a few Apps 
4. Try to save the profile 
5. The app will either 
	A. Stop responding 
	B. The screen becomes fully white for a fraction of a second and go back to main menu
6. In both cases the Apps VPN Mode setting is not saved  

**Expected behavior**
Apps VPN Mode setting should be saved normally without error 

**Screenshots**
![Screenshot_20190608-222429](https://user-images.githubusercontent.com/5536733/59151193-db665400-8a3f-11e9-82b5-ac63b5aeebf7.jpg)

**Smartphone**
 - Android/Chrome OS version: Android 9 Build PQ3A.190605.003 June 5, 2019 Patch 
 - Device: Google Pixel 3
 - Version: v4.8.0
 - Last version that did not exhibit the issue: 

**Configuration**
Put an `x` inside the [ ] that applies.

* [ ] IPv4 server address
* [ ] IPv6 server address
* [ ] Client IPv4 availability
* [ ] Client IPv6 availability
* Encrypt method:
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [ ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [X] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS over UDP
* Plugin configuration (if applicable):
* [ ] Auto Connect
* [ ] TCP Fast Open
* If you're not using VPN mode, please supply more details here:

**Additional context**
I attached the phone to logcat, and got the following 

`06-08 21:22:12.905 20731 20731 W ActivityThread: handleWindowVisibility: no activity for token android.os.BinderProxy@c15f198
06-08 21:22:14.914 20731 20731 D AndroidRuntime: Shutting down VM
06-08 21:22:14.915 20731 20731 E AndroidRuntime: FATAL EXCEPTION: main
06-08 21:22:14.915 20731 20731 E AndroidRuntime: Process: com.github.shadowsocks, PID: 20731
[...]

PR #2236, opened 2019-06-09
Fix: #2234

huanght1997:


madeye:
Looks a valid fix. 

@Mygod What do you think?

Mygod:
I don't think this has anything to do with #2234.

PR #2263, opened 2019-07-19
Immersive theme: Make navigation bars translucent

Mygod:
TODO/Wish list:

* [ ] Main layout navbar color is currently hacky. Probably should fix this. Alternatively, we can make main layout even more translucent.
* [ ] Should we remove background dim on Android Q for [gestural navigation](https://developer.android.com/preview/features/gesturalnav)?


Mygod:
Alright let's take another look at this when Android Q is officially released.

PR #1027, opened 2016-12-24
New UI with drawer (#762)

Mygod:
I have finished implementing new things for this new UI. What remains to be done is to fix these things now. DON'T MERGE BEFORE FINISHING THESE.

* [x] New banner? (assigned to @madeye)
* [x] License information in new files; (`/.headache.cfg`) (assigned to @madeye)
* [x] More thorough tests! (for example, editing should be disabled when connecting/stopping, etc.) (assigned to @Mygod)
* [x] Test & fix tasker features; (assigned to @Mygod)
* [x] Localizations; (assigned to @madeye)
* [x] Remove resources that are no longer used; (assigned to @Mygod)
* [x] Final review by @Mygod. (because I can't add myself to the reviewers :disappointed:)

P.S. Snackbar animation is broken, but that isn't a big issue for now. :stuck_out_tongue: Oh and please don't squash and merge for this PR.

After finishing this, we can start working on #1006 #1009 #1021. :tada:

PR #2296, opened 2019-08-23
Show stats bar only when connected

Mygod:
Finally I finish the original vision in #1027.

Slight issue: You can see stats bar sliding down after `onCreate`.


Mygod:
@madeye thoughts?

madeye:
LGTM.

PR #2306, opened 2019-09-12
First phase cleaning the project lint warnings 

VMadalin:
### Description
Several cleaning of project lint warnings on all modules, add lint validation as a task on `circleCI`(_**./gradlew lintDebug**_) for maintain the code quality in future PR

### Overview
- Core module:
Screenshot 2019-09-10 at 23 31 56
- Mobile module:
Screenshot 2019-09-12 at 03 13 35
- Plugin module:
Screenshot 2019-09-12 at 03 14 40
- TV module:
Screenshot 2019-09-12 at 03 15 51

### Next Steps
This is just a small step for project maintenance, and it need to continue reviewing code style, performance, some suppress warnings, memory leaks and more

PR #2308, opened 2019-09-17
Update dependencies

Mygod:
Fixes #2306.

Warning: While errors are unlikely, this PR is untested because I do not have time.

@VMadalin For the rest of your changes, there are good reasons (to me) to ignore the warnings and not suppress them at the same time.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         minSdkVersion = 21
         sdkVersion = 29
         compileSdkVersion = 29
-        lifecycleVersion = '2.2.0-alpha03'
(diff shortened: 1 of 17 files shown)


VMadalin:
Thank you for make the `patch` of some change from my PR and added new one. I understand your point of view to ignore the rest of warnings for the moment, and don't tacking them into count. But remember some of them are more or less important like the following, and they should not be forgotten.

```
Internationalization:Bidirectional Text
--
2 | warning RtlHardcoded: Using left/right instead of start/end attributes
1 | warning RtlEnabled: Using RTL attributes without enabling RTL support
```
This break the layout for arabic languages.

PR #2378, opened 2019-12-10
Match disabled apps and update sort method

shadowsocksRb:
对 #2335 的后续补充

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/AppManager.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/AppManager.kt
@@ -31,6 +31,9 @@ import android.content.Intent
 import android.content.pm.ApplicationInfo
 import android.content.pm.PackageInfo
 import android.content.pm.PackageManager
+import android.content.pm.PackageManager.GET_PERMISSIONS
+import android.content.pm.PackageManager.MATCH_DISABLED_COMPONENTS
+import android.content.pm.PackageManager.MATCH_UNINSTALLED_PACKAGES
 import android.graphics.drawable.Drawable
 import android.os.Bundle
 import android.util.SparseBooleanArray
@@ -79,7 +82,7 @@ class AppManager : AppCompatActivity() {
             }
             // Labels and icons can change on configuration (locale, etc.) changes, therefore they are not cached.
             val cachedApps = cachedApps ?: pm.getInstalledPackages(
-                    PackageManager.GET_PERMISSIONS or PackageManager.MATCH_UNINSTALLED_PACKAGES)
+                    GET_PERMISSIONS or MATCH_UNINSTALLED_PACKAGES or MATCH_DISABLED_COMPONENTS)
                     .filter {
                         when (it.packageName) {
                             app.packageName -> false
@@ -93,7 +96,7 @@ class AppManager : AppCompatActivity() {
         }
     }
 
-    private class ProxiedApp(private val pm: PackageManager, private val appInfo: ApplicationInfo,
+    private class ProxiedApp(private val pm: PackageManager, val appInfo: ApplicationInfo,
                              val packageName: String) {
         val name: CharSequence = appInfo.loadLabel(pm)    // cached for sorting
         val icon: Drawable get() = appInfo.loadIcon(pm)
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/com/github/shadowsocks/AppManager.kt
```diff
@@ -79,7 +82,7 @@ class AppManager : AppCompatActivity() {
             }
             // Labels and icons can change on configuration (locale, etc.) changes, therefore they are not cached.
             val cachedApps = cachedApps ?: pm.getInstalledPackages(
-                    PackageManager.GET_PERMISSIONS or PackageManager.MATCH_UNINSTALLED_PACKAGES)
+                    GET_PERMISSIONS or MATCH_UNINSTALLED_PACKAGES or MATCH_DISABLED_COMPONENTS)
```
Mygod: Why is matching components useful for getting packages?

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
```diff
@@ -135,7 +138,11 @@ class AppManager : AppCompatActivity() {
             apps = getCachedApps(packageManager).map { (packageName, packageInfo) ->
                 coroutineContext[Job]!!.ensureActive()
                 ProxiedApp(packageManager, packageInfo.applicationInfo, packageName)
-            }.sortedWith(compareBy({ !isProxiedApp(it) }, { it.name.toString() }))
+            }.sortedWith(compareBy(
+                    { !isProxiedApp(it) },
+                    { it.appInfo.flags and ApplicationInfo.FLAG_SYSTEM },
+                    { it.name.toString() }
+            ))
```
Mygod: I consider this change to be confusing to users. Please do not put two (or more) unrelated changes in a single pull request in the future.

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
shadowsocksRb: AOSP没有`MATCH_DISABLED_PACKAGES`，这个flag用在这里就是匹配已停用应用。
这个问题需要问Google才有答案

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
shadowsocksRb: 还有这种理由？！用户会疑惑为什么某某应用没有显示，括号里的数字是什么，为什么有些数字一样。但唯独不在乎排序乱了：能找到就行，快点更好。

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
Mygod: You should provide relevant documentations/blog posts/Android platform source code to back up your claim. You should not expect me to read through the source code of `PackageManagerService.java` to find out whether this flag is indeed useful, or it is just a coding mistake someone else made online and you just copied their code mindlessly.

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
Mygod: I am still unconvinced as there are also numerous other sorting options that are still as valid *and* as arbitrary as this. By Occam's razor I prefer the old way.

@madeye can weigh in on this.

madeye: We should avoid this change.

It's intuitive that we put selected apps on the top. However, if we keep adding other sorting options here, then users will get confused easily.

mobile/src/main/java/com/github/shadowsocks/AppManager.kt
shadowsocksRb: https://android.googlesource.com/platform/frameworks/base/+/a9ef0e9bb4df194ea5a2575c8ef0fe8d1572b1f4/core/java/android/content/pm/PackageManager.java#176
同样讨厌去pm service里寻找具体逻辑。这个可以直接在AS跳转源码里看见，`ApplicationInfoFlags`注解里包括了`MATCH_DISABLED_COMPONENTS`，这应该很明显。

Mygod: `getInstalledPackages` returns a list of `PackageInfo` that can contain other `ComponentInfo`. The documentation of `MATCH_DISABLED_COMPONENTS` reads `flag: include disabled components in the returned info`. As far as I see, this flag serves no purpose for this.

shadowsocksRb: …………
这些代码已经在rb里用上了，行为符合预期，一切良好。

Mygod:
I stand with my original decisions in #2335. Just a reminder that *please* do not put two (or more) unrelated changes in a single pull request in the future. Having one change for every pull request would make the reviewing much easier and we can merge approved changes much faster. Thanks again.

PR #2386, opened 2019-12-13
Support matching IPv6 in LocalDnsServer

Mygod:
This fixes confusions once and for all in like fa3d4e78b8104c98812802f9aa8ba163fe59f0d7.

Also updates China IP address blocks.

P.S. I am not sure what is the most authentic source for IP address blocks. If it is not ipverse.net, feel free to change it. 😛


Mygod:
P.S. Tested locally and it bypasses baidu.com correctly but not ip.cn. For IPv6, tested ipv6.google.com and ipv6.jmu.edu.cn and works as expected except that Chrome really does not seem to like NAT6 addresses.

Mygod:
@madeye We should probably add these updates to `bypass-china.acl` and `bypass-lan-china.acl` as well. What do you think if we just extract bypass IPs from ACL file?

madeye:
Yeah, extracting bypass IPs from ACL files should be better

Mygod:
@madeye Done. It seems to be working alright but now start up takes ~1s extra.

Mygod:
Hostname matching in `LocalDnsServer` might also add ~80ms of delay per DNS request. Unfortunately the pattern matching also creates a lot of garbage objects for every DNS request.

madeye:
Is it possible that we define a special section, e.g. `[china_ip_range]` in ACL? If so we can still use the raw ip range?

Mygod:
Oh I meant that I also implemented the full hostname matching in `LocalDnsServer` which was not there before. The worse part about this is a lot of `Matcher` objects are being created so the garbage collector will kick in very frequently if there are many DNS queries which could cause battery drain. If that part is removed the performance should be pretty good.

For the start up performance hit, profiling shows that the ~1 second comes mainly from `SortedList` in `Acl`.

madeye:
I see. 

BTW, can we try some matcher like https://github.com/google/re2j and compile all rules into one `a|b|c|d...`? Then we can keep only one matcher during the whole lifetime of a VPN session.

Mygod:
That might work but we need to remove almost all groups  in our regex, otherwise it is still going to use a lot of memory. Currently almost every rule has `(^|\.)` at the beginning.

Also we probably also need to optimize `Acl`.

madeye:
Let's make the improvement a feature work.

IMO, for domain ACL, wildcard like rules should work for almost all the cases.

Mygod:
Good news: now hostname matching takes <1ms.
Bad news: Initialization takes 3s. :)

Mygod:
If we only allow googleapi rules as before, this change causes start up to be only 16x slower. (reading acl is taking the majority of the time) We can mitigate start up slowdown by putting it into async.

Mygod:
Did some more testing. On average, ACL set up time is ~400ms, hostname lookup is <1ms (~50ms if we use full ACL), IPv4 subnet lookup ~2ms and IPv6 subnet lookup <1ms. While these time is negligible in terms of user experience, full ACL lookup can accumulate into battery drain since it is ~50ms of CPU time for every query.

madeye:
Cool!

madeye:
I guess we can merge this now? @Mygod

Mygod:
There's still some cleanup to be done. Sorry. :)

madeye:
No hurry...

PR #2388, opened 2019-12-23
Move banner from mipmap to drawable

mahc9kez:
I saw this warning when I open `core/src/main/res/drawable-anydpi-v24/banner.xml` in Android Studio

> <vector> XML file should be in "drawable", not "mipmap"

https://developer.android.com/guide/topics/manifest/application-element

> android:banner
> A drawable resource providing an extended graphical banner for its associated item.


mahc9kez:
huh may I know why this was closed? seems mipmap should only be used for saving launcher icons, saving banner there result in Android Studio refuse to preview it.

PR #2402, opened 2020-01-11
Refactor subscriptions

Mygod:
Probably finish up tonight.


madeye:
Cool, hope we can have a pre-release next week.

Mygod:
@madeye Hmm wouldn't it make more sense to simply use an EditText for all the subscription URLs?

madeye:
I think it's much better to have a fragment to manage all subscriptions...

Mygod:
EditText allows for copy/paste batch actions though.

madeye:
One concern is that it would be hard to validate each URL, like reporting insecure URLs.

Given we already have the subscription fragment, I don't think we should remove it. At least, we can extend subscription based on this for future development, e.g. subscription traffic statics.

Mygod:
Okay. Should we extend this feature to tv?

madeye:
On TV, I think one EditText should be good enough.

With subscription support on TV, I think we can improve the Google Play score a lot...

Mygod:
Alright. Not sure what future extensions you are planning for but I will try to keep changes minimal. :)

Mygod:
Hmm I'm currently getting `java.lang.VerifyError: Verifier rejected class com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1: java.lang.Object com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1.invokeSuspend(java.lang.Object) failed to verify: java.lang.Object com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1.invokeSuspend(java.lang.Object): [0x14A] 'this' argument 'Uninitialized Reference: com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1$4 Allocation PC: 328' not instance of 'Precise Reference: com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1$2' (declaration of 'com.github.shadowsocks.subscription.SubscriptionService$onStartCommand$1$invokeSuspend$$inlined$map$lambda$1' appears in /data/app/com.github.shadowsocks-MdKJd5Xmv6syOmQxl84ysw==/base.apk!classes4.dex)`. Seems like compiler bug.

Mygod:
I cannot be bothered to make a minimal example to reproduce this error and report to Kotlin.

@madeye Can you help test this?

madeye:
Sure.

perqin:
Will libev port implement this subscription mechanism? I am wondering how this feature will be implemented by those ports without GUI. I am using libev port on my Linux desktop, and it will be nice to have subscription support.

PR #2403, opened 2020-01-12
Trust signature of V2ray Plugin F-Droid build

mahc9kez:
Signature extracted from apk which downloaded from https://f-droid.org/en/packages/com.github.shadowsocks.plugin.v2ray/

`keytool -printcert -jarfile ./com.github.shadowsocks.plugin.v2ray_1030000.apk -rfc`
or
`unzip -p ./com.github.shadowsocks.plugin.v2ray_1030000.apk META-INF/EABA26EE.RSA | keytool -printcert -rfc`

### DIFF
--- a/core/src/main/java/com/github/shadowsocks/plugin/PluginManager.kt
+++ b/core/src/main/java/com/github/shadowsocks/plugin/PluginManager.kt
@@ -96,6 +96,28 @@ object PluginManager {
                     |Iqonxpwk2ay+Dm5RhFfZyG9SatM/JNFx2OdErU16WzuK1ItotXGVJaxCZv3u/tTwM5aaMACGED5n
                     |AvHaDGCWynY74oDAopM4liF/yLe1wmZDu6Zo/7fXrH+T03LBgj2fcIkUfN1AA4dvnBo8XWAm9VrI
                     |1iNuLIssdhDz3IL9Yg==
+                  """, Base64.DEFAULT)) +
+                Signature(Base64.decode( // FDroid (V2ray Plugin)
+                """
+                    |MIIDXzCCAkegAwIBAgIEY71mbDANBgkqhkiG9w0BAQsFADBgMQswCQYDVQQGEwJV
+                    |SzEMMAoGA1UECBMDT1JHMQwwCgYDVQQHEwNPUkcxEzARBgNVBAoTCmZkcm9pZC5v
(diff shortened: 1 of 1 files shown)


Mygod:
How does one ensure that F-Droid will use different signatures for each app?

Just for the records, I also have mixed feelings regarding F-Droid.

mahc9kez:
idk, just checked documentation

> **APK Signing**
> The F-Droid tools can automatic generate and manage APK signing keys for each app that it maintains. With the repo index signing configured, all that remains to be done for package signing to work is to set the keydname field in config.py to contain the same identifying details you entered before. A new key will be generated using these details, for each application that is built. If a specific key is required for a particular application, this system can be overridden using the keyaliases config settings.

https://f-droid.org/en/docs/Signing_Process/

```
# Use this to override the auto-generated key aliases with specific ones
# for particular applications. Normally, just leave it empty.
# keyaliases = {}
# keyaliases['com.example.app'] = 'example'
# You can also force an app to use the same key alias as another one, using
# the @ prefix.
# keyaliases['com.example.another.plugin'] = '@com.example.another'
```
https://gitlab.com/fdroid/fdroidserver/blob/master/examples/config.py

But idk if the `config.py` that F-Droid server used to build apps are publicly available. I'm asking F-Droid contributors.

mahc9kez:
I figured out a solution, instead of get the config we can get the sha256 fingerprint of APKs which can be installed or downloaded from F-Droid.

https://gitlab.com/fdroid/fdroiddata/merge_requests/6105#note_270339704

Just open https://f-droid.org/repo/index.xml (or index.jar or index-v1.jar which are signed) in a browser, search the package name, copy and search the its value of `sig` property to find APKs which signed with the same key.

madeye:
I think currently we didn't reject any plugin with untrusted signature, right?

madeye (commented):
I don't think we should trust any third-party signature, even it's from F-droid.

Mygod:
@madeye Yes and agreed.

mahc9kez:
I thought the f-droid build of v2ray-plugin could be considered as official release, it just feels a bit strange that when the app and plugin are official but it still says it's untrusted. But yeah I saw the point, it's just a warning and plugin still being loaded and functional, and it only effect ppl who use f-droid build of v2ray-plugin with the official app.

UjuiUjuMandan:
https://f-droid.org/docs/Reproducible_Builds/ "Exclusively publishing (upstream) developer-signed APKs"

PR #2431, opened 2020-02-04
Plugin 1.3.4

Mygod:
Main change: support plugin aliases proposed by @CzBiX: https://github.com/shadowsocks/v2ray-plugin-android/issues/18#issuecomment-581262267
For other changes, see CHANGES.md.

I have not thoroughly tested this. @madeye Can you help? (client app should not require a plugin lib update)

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/ProfileConfigFragment.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/ProfileConfigFragment.kt
@@ -52,6 +52,7 @@ class ProfileConfigFragment : PreferenceFragmentCompat(),
 
         private const val REQUEST_CODE_PLUGIN_CONFIGURE = 1
         const val REQUEST_UNSAVED_CHANGES = 2
+        private const val REQUEST_PICK_PLUGIN = 3
     }
 
     @Parcelize
@@ -69,7 +70,7 @@ class ProfileConfigFragment : PreferenceFragmentCompat(),
 
     private var profileId = -1L
     private lateinit var isProxyApps: SwitchPreference
-    private lateinit var plugin: IconListPreference
+    private lateinit var plugin: PluginPreference
     private lateinit var pluginConfigure: EditTextPreference
     private lateinit var pluginConfiguration: PluginConfiguration
     private lateinit var receiver: BroadcastReceiver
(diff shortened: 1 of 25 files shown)


madeye:
To test this change, I think we need to update the plugin app accordingly?

Mygod:
Done.

Mygod:
To test 4a6dc8049c71f537059215685700ba5e33baf633, we could test against [GoQuiet-android](https://github.com/cbeuw/GoQuiet-android).

madeye:
I tested it locally and found one issue.

In the profile fragment, after switching the plugin, the option didn't get updated. Then saving the profile and open the profile again, the plugin option is updated.

mobile/src/main/java/com/github/shadowsocks/ProfileConfigFragment.kt
```diff
@@ -102,18 +103,6 @@ class ProfileConfigFragment : PreferenceFragmentCompat(),
         findPreference<Preference>(Key.udpdns)!!.isEnabled = serviceMode != Key.modeProxy
         plugin = findPreference(Key.plugin)!!
         pluginConfigure = findPreference(Key.pluginConfigure)!!
-        plugin.unknownValueSummary = getString(R.string.plugin_unknown)
-        plugin.setOnPreferenceChangeListener { _, newValue ->
```
madeye: Why removed `setOnPreferenceChangeListener` here?

Mygod: Hmm it is supposed to be handled by `onActivityResult` now.

madeye: Do you where the plugin icon and name updated when the plugin is switch in the dialog? I'm debugging the plugin switching issue.

mobile/src/main/java/com/github/shadowsocks/ProfileConfigFragment.kt
```diff
@@ -221,6 +206,19 @@ class ProfileConfigFragment : PreferenceFragmentCompat(),
...
+                val override = pluginConfiguration.pluginsOptions.keys.firstOrNull {
+                    plugin.plugins.lookup[it] == selected
+                }
+                pluginConfiguration = PluginConfiguration(pluginConfiguration.pluginsOptions, override ?: selected.id)
+                DataStore.plugin = pluginConfiguration.toString()
+                DataStore.dirty = true
+                pluginConfigure.isEnabled = selected !is NoPlugin
+                pluginConfigure.text = pluginConfiguration.getOptions().toString()
+                if (!selected.trusted) Snackbar.make(view!!, R.string.plugin_untrusted, Snackbar.LENGTH_LONG).show()
+            }
```
Mygod: @madeye This is supposed to be the callback.

mobile/src/main/java/com/github/shadowsocks/ProfileConfigFragment.kt
Mygod: Committed a fix via web editor. Let me know if it works. :)

Mygod:
Cool. Let's merge this for the next release.

Issue #2452, opened 2020-02-20
Considering alternative coroutine sslocal impls


Mygod:
Moving discussion from https://github.com/shadowsocks/shadowsocks-org/issues/154.

We are trying to make ACL handling of shadowsocks-android less hacky (for example https://github.com/shadowsocks/shadowsocks-libev/issues/2627), however, things seem challenging.

1. While shadowsocks-libev more or less takes the approach of doing async I/O via libev, libev requires you to manually keep track of your execution point to resume your connection processing. This makes extending functionalities error-prone and overall the code is hard to read. In fact, there are still traces of synchronous I/O in shadowsocks-libev (e.g. `get_sockaddr`), which could block the entire thread if network conditions are bad.
2. Modern programming language more or less solves this problem with coroutines, examples include golang/rust/c++20/Kotlin. However, go-shadowsocks2 (missed opportunity for go2shadowsocks btw) and shadowsocks-rust do not have support for ACL files yet. As we use really large ACL files, we need to integrate re2 or look for a regex library with similar performance. (golang stdlib does not suffice: https://github.com/golang/go/issues/11646)
3. Implementing everything in JVM is a bad idea since every byte array is going to be an object in GC heap, and there are no good Kotlin async I/O libraries either. It seems like there are similar issues in golang (objects escaping to heap) but I am not an expert in golang.
4. Doing PAC is going to be even more expensive than ACL since PAC is actually Javascript.

Thoughts are appreciated.

CC @madeye @riobard @zonyitoo

riobard:
For point 2, Go's RE engine is pretty [robust](https://swtch.com/~rsc/regexp/regexp1.html).

For point 3, Go's GC is really good now and it's relatively easy to to write memory-efficient code. But compared to languages without GC and runtime (Rust/C/C++) there's still a memory penalty. However [recent discussion on HN](https://news.ycombinator.com/item?id=22336284) makes me more confident in Go compared to Rust. 


@Mygod I'm like the go2shadowsocks name! 😂

Mygod:
I don't know that article but it seems ancient (it even predates re2). We are talking about 100KB of regex pattern matching domain names so I am not sure if it is as performant as you claimed.

Some benchmarks would be nice...

riobard:
The point of Cox's article is that Go uses finite automata-based regex engine so the performance should be good enough. Just checked re2 is similar so I guess you should expect similar performance.

Mygod:
[...]

PR #2157, opened 2019-03-14
support hosts

ayanamist:
fix #2150

PR #2454, opened 2020-02-23
Integrate shadowsocks-rust

madeye:
This PR aims to replace shadowsocks-libev with shadowsocks-rust.

Fix issue #2452.

- [x] Gradle and Cargo integration
- [x] Build pass
- [x] VPNService callback
- [x] TCP relay
- [x] DNS realy
- [x] UDP relay
- [x] ACL (IP rules)
- [x] ACL (domain name rules)
- [x] Traffic statistics

Nice to have:
- [x] Parallel lookup to local and remote
- [x] Unix domain socket to local DNS resolver.

Future development plan (maybe not covered by this PR)
- [ ] A local DNS cache
- [ ] Local DNS resolver in Rust
- [ ] Fine-grained policy routing
- [ ] Multiple upstreams


madeye:
After `./gradlew cargoBuild`, we will get executables like `core/shadowsocks-rust/target/armv7-linux-androideabi/release/sslocal`.

Can we write a gradle task to move them into jni folder? @Mygod

Mygod:
Hmm shall we put source code in folder `core/src/main/jni` or `core/src/main/rust`?

Shouldn't Gradle plugin take care of the binary bundling?

madeye:
Yeah, the plugin put those executables here: ./core/build/rustJniLibs/android/armeabi-v7a/sslocal

But I'm not very sure if it will be packed in APK and installed correctly as it's an executable.

Mygod:
Only one way to find out. :)

madeye:
The naming issue should be fixed via a51ad37.

madeye:
I've finished the `protect()` callback and get the TCP relay, UDP relay, ACL (IP rules) works now.

Further development:

1. Traffic stat requires changes in shadowsocks-rust's local client. I have already added the stat_path RPC interface. @zonyitoo can help to implement the traffic stat logic in local client.
2. ACL (domain name rules) requires reverse local DNS lookup. @Mygod can help with this part. It requires our local DNS resolver supports `PTR` queries for `in-addr.arpa` addresses. We may need a IP/domain map in our local DNS resolver for this feature.

zonyitoo:
> Traffic stat requires changes in shadowsocks-rust's local client. I have already added the stat_path IPC interface. @zonyitoo can help to implement the traffic stat logic in local client.

Easy piece. I will try to do that in this weekend.

Mygod:
@madeye In my proposal, I was saying that there is no need for a reverse DNS lookup -- this can be easily done using a reverse lookup map in `sslocal`. I think this is the better way as it also saves a RPC.

madeye:
@Mygod Okay, if so, we need to implement the local resolver in sslocal. Can you help on this?

I think it can be implemented based on trust-dns. The good news is shadowsocks-rust has already used it for safe DNS resolving. What we need is to extend it to a ChinaDNS like local resolver.

Mygod:
I could if y'all can wait for me to pick up rust first. :)

Mygod:
Oh I was thinking of putting the code at `core/src/main/rust/shadowsocks-rust` and we can put other ancillary code at the rust directory... Let me see if I will have time to work on this over the weekends.

core/src/main/res/values/arrays.xml
```diff
@@ -1,44 +1,16 @@
...
-        <item>aes-192-ctr</item>
-        <item>aes-256-ctr</item>
-        <item>bf-cfb</item>
-        <item>camellia-128-cfb</item>
-        <item>camellia-192-cfb</item>
-        <item>camellia-256-cfb</item>
-        <item>salsa20</item>
-        <item>chacha20</item>
-        <item>chacha20-ietf</item>
+        <item>plain</item>
```
Mygod: We might want to keep the old ciphers for backwards compatibility...

madeye: Integrating openssl is much difficult than I thought. Is it acceptable to use prebuilt binaries?

Mygod: binaries for openssl? We can definitely use it as a WIP and worry about it later.

madeye: OpenSSL also works now by enabling `opnessl-vendored` feature.

I didn't expect we can integrate openssl in that way... It's really cool.

Mygod: Isn't modern dependency management amazing? (except when they don't work)

zonyitoo:
It seems that adding a flow stat for `sslocal` is far more complicated than I was expected. :(

madeye:
No hurry... In shadowsocks-libev, I have a counter in the context and a timer to push the counter value to shadowsocks-android.

zonyitoo:
Because `shadowsocks-rust` could be run as a library, so I have to manage these stat variables by using reference counter.

Mygod:
Well it looks like this repo is going to take more than 4GB of RAM to build. Let me try again tomorrow. :)

zonyitoo:
> Well it looks like this repo is going to take more than 4GB of RAM to build. Let me try again tomorrow. :)

Consider disable LTO. It takes quite a lot of memory while linking.

build.gradle
```diff
@@ -14,6 +14,7 @@ buildscript {
         junitVersion = '4.13'
         androidTestVersion = '1.2.0'
         androidEspressoVersion = '3.2.0'
+        ndkVersion = '21.0.6113669'
```
Mygod: Why do we need to specify ndkVersion btw?

madeye: I see the warning message that `ANDROID_NDK_HOME` is deprecated now. Any suggestion?

Mygod: I don't think we are using that env?

core/build.gradle
```diff
@@ -44,12 +45,29 @@ android {
...
+    libname = 'sslocal'
+    targets = ['arm', 'arm64', 'x86', 'x86_64']
+    profile = 'debug'
+    targetIncludes = ['libsslocal.so']
+    extraCargoBuildArguments = ['--bin', 'sslocal']
+    features {
+        noDefaultBut "sodium", "android", "rc4", "aes-cfb", "aes-ctr", "camellia-cfb", "openssl-vendored"
+    }
+    exec { spec, toolchain ->
+        spec.environment("RUSTFLAGS", "-C link-arg=-o -C link-arg=target/${toolchain.target}/${profile}/libsslocal.so")
```
Mygod: I would think that the plugin would take care of the output so name... Is this an upstream issue?

build.gradle
madeye: Otherwise, I cannot get the project built with Android Studio 4.0...

core/build.gradle
madeye: I think it's actually a limitation of Cargo, which cannot customize the output binary name.

Mygod: Their sample app is far less involved than this. I would assume it's upstream bug (I wonder if anyone is actually using the plugin).

madeye: I got the workaround here: https://github.com/rust-lang/cargo/issues/1706#issuecomment-563889929

Mygod: I think they already have a workaround: https://github.com/mozilla/rust-android-gradle/blob/3dc5511b2bbed4b66d03e360302b8147d7138877/plugin/src/main/kotlin/com/nishtahir/CargoBuildTask.kt#L193

Mygod: https://github.com/mozilla/rust-android-gradle/issues/30

madeye:
I think the submodule is not correctly moved in this branch...

madeye:
The latest commit should fix the issue.

Mygod:
Yeah sorry for force pushing lol. @madeye

Mygod:
Sadly it seems like I don't have time to understand the rust part for now, so instead I just added a JVM part for `local_dns_path`. You can use it just like a nonreusable TCP DNS (DNS request/response format `<length:i16><payload>`). The implementation is extremely drafty for now but it will serve its purpose for the time being. Thanks guys for the hard work! 👍

Mygod:
P.S. It is only enabled for `VpnService` for now because I am not sure what is going on in shadowsocks-rust for now.

There is a possibly more efficient implementation -- we can let `local_dns_path` instead simply return the desired `netId: net_handle_t` and then shadowsocks-rust can handle the DNS resolving using [Android NDK APIs `android_getaddrinfofornetwork` or `android_res_nsend`](https://developer.android.com/ndk/reference/group/networking).

madeye:
Currently, there's no DNS queries from shadowsocks-rust, as the SOCKS5 requests from tun2socks are all IPv4 or IPv6 types, no domain name. That's also why ACL (domain name) doesn't work now.

With the `local_dns_path`, I will try to implement an internal reverse DNS resolver to enable ACL (domain name).

Mygod:
@madeye No... I am thinking that `sslocal` will listen on port 5450 (or local DNS port) directly to handle the DNS traffic, and use `local_dns_path` to do local DNS lookups. I want to minimize the packet handling in JNI and use Rust as much as possible. :)

Mygod:
I would like to propose the following to-do list (based on the one in OP), with priorities.

* ACL function
* Traffic statistics (assigned to @zonyitoo?)
* Move `LocalDnsServer` implementation to rust using `local_dns_path` (but I guess we could start with using `LocalDnsServer` for all the work, and change `local_dns_path` to use `LocalDnsServer` instead? but then we need to write some code that we will later throw away)

Lower priority:
* Investigate the easiest way to do DNS resolving in rust in proxy/transproxy mode (in these mode we always send our traffic to default network. should we use `local_dns_path` vs some native approach in rust?)
* Investigate possibility of using `android_res_nsend` directly in sslocal? https://github.com/shadowsocks/shadowsocks-android/pull/2454#issuecomment-593045810
* Implement async I/O for LocalSocket in JNI (assigned to @Mygod)
* Investigate binary size and make rust binary smaller?
* Code review & cleanup after we get things working

Lowest priority: (these are usually new features, can implement these in the future)
* Fine-grained policy routing
* Rust tun2socks? (https://github.com/iCepa/tun2tor)
* If we have multiple rust binary, maybe we can make a master binary so that stdlib is shared, etc
* Make tun2socks pass source address so that we can look up connection by UID (useful for firewall, etc)
* Multiple upstreams (I wonder what an intuitive user interface would look like for this)

Thanks everyone again for the hard work!

.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           key: jars-{{ checksum "build.gradle" }}
       - run:
           name: Run Build and Tests
-          command: ./gradlew assembleDebug check
+          command: ./gradlew cargoBuild && ./gradlew assembleDebug check
```
Mygod: This change is no longer needed after my patch.

core/src/main/java/com/github/shadowsocks/bg/BaseService.kt
```diff
@@ -232,7 +232,7 @@ object BaseService {
             }
         }
 
-        fun buildAdditionalArguments(cmd: ArrayList<String>): ArrayList<String> = cmd
+        fun isVpnService(): Boolean = false
```
Mygod: The more Kotlin way to write this is `val isVpnService get() = false`. 😝

core/src/main/java/com/github/shadowsocks/bg/ProxyInstance.kt
```diff
@@ -114,27 +114,25 @@ class ProxyInstance(val profile: Profile, private val route: String = profile.ro
...
+        val cmd = arrayListOf(
                 File((service as Context).applicationInfo.nativeLibraryDir, Executable.SS_LOCAL).absolutePath,
-                "-b", DataStore.listenAddress,
-                "-l", DataStore.portProxy.toString(),
-                "-t", "600",
-                "-S", stat.absolutePath,
-                "-c", configFile.absolutePath))
+                "--stat-path", stat.absolutePath,
+                "-c", configFile.absolutePath)
+        if (service.isVpnService()) cmd += arrayListOf("--protect-path", "protect_path")
```
Mygod: I think we hardcoded the paths `./protect_path` and `./stats_path` in shadowsocks-libev. I don't think it would be useful to have an option for customization here...

zonyitoo:
> Traffic statistics (assigned to @zonyitoo?)

I could make a simple solution first. Some refactor is needed.

Mygod:
Again as @madeye said no hurry...

madeye:
@Mygod Feel free to change anything here...

core/src/main/java/com/github/shadowsocks/bg/ProxyInstance.kt
madeye: Okay, let's use hardcoded path in shadowsocks-rust as well.

zonyitoo:
@madeye I have already implemented traffic statistic. Please pull shadowsocks-rust's master.

madeye:
Traffic stat works now. Cool!

zonyitoo:
> Investigate binary size and make rust binary smaller? (say https://github.com/johnthagen/min-sized-rust?)

Consider disable `trust-dns` feature. It will significantly shrink the output binary's size. BUT `dns_resolve()` call will be slower.

Mygod:
Don't worry about binary size too much for now. Furthermore, it is better to save CPU cycles/battery than a few megabytes of storage on modern mobile. ;)

madeye:
@zonyitoo @Mygod I built a simple China DNS here: https://github.com/madeye/trust-china-dns. I think later we can integrate it into shadowsocks-rust.

zonyitoo:
I couldn't see any logic code in this repo. Have you pushed your commits?

madeye:
@zonyitoo Yeah, I put the logic code in `bin/main.rs`... https://github.com/madeye/trust-china-dns/blob/master/bin/main.rs

madeye:
Opened a pull request for DNS relay: https://github.com/shadowsocks/shadowsocks-rust/pull/202

core/build.gradle
Mygod: Yeah as I figured nobody is actually maintaining this plugin.

madeye:
All high priority tasks finished now. @Mygod maybe you can continue to work on some refinements.

Mygod:
Cool! Let me get around to it...

Mygod:
@madeye Are you making any changes currently? I can work on this a bit in the next few hours.

Mygod:
Ah feel free to make changes I guess I am still reading... :)

madeye:
Almost done here.

I think you can refine the UNIX domain name local DNS first. Then figure out a way to process hosts file in rust...

zonyitoo:
> Then figure out a way to process hosts file in rust...

https://docs.rs/trust-dns-resolver/0.19.3/trust_dns_resolver/struct.Hosts.html

Mygod:
Well we can just remove the hosts file feature if it's too complicated and let whoever wants to use it (#2157) implement it. :)

zonyitoo:
What is `dns-relay` for? Is it a DNS proxy like ChinaDNS? If so, what is the `reverse_lookup_cache` for?

Mygod:
Yes it is DNS proxy plus DNS for socksifier. I think that is the `exceptionIpPool` that I explained [here](https://github.com/shadowsocks/shadowsocks-android/issues/2452#issuecomment-589503820).

zonyitoo:
I have merged all clients into one `sslocal` binary.

But what's the status of this PR? I am still waiting for releasing them all in v1.8.10.

madeye:
@zonyitoo This PR is blocked by this one: https://github.com/shadowsocks/shadowsocks-rust/pull/211

However, all functions are working now in shadowsocks-android and shadowsocks-rust. The only left issue is if @Mygod plans to do some refactoring to move implementations from `dnsrelay/mod.rs` to `dnsrelay/upstream.rs`.

If not, I can help to remove `dnsrelay/upstream.rs`, which should unblock all of us.

Mygod:
Yeah busy working on other things currently (and will be for at least another week probably) but I don't think this should block release schedule of shadowsocks-rust?

zonyitoo:
There is no schedule. :P  Just take your time.

Mygod:
So I guess we can work with shadowsocks-rust at its current state and we can delay https://github.com/shadowsocks/shadowsocks-rust/pull/211... I am not sure if local dns path works properly currently, as I am getting a lot of "Borken pipe" warnings.

.gitmodules
```diff
@@ -17,19 +13,6 @@
...
-	branch = stable
-[submodule "core/src/main/jni/libev"]
-	path = core/src/main/jni/libev
-	url = https://git.lighttpd.net/libev.git
-[submodule "core/src/main/jni/re2"]
-	path = core/src/main/jni/re2
-	url = https://github.com/google/re2.git
+[submodule "core/src/main/rust/shadowsocks-rust"]
+	path = core/src/main/rust/shadowsocks-rust
+	url = https://github.com/madeye/shadowsocks-rust
```
Mygod, line 19: Should this be changed to `shadowsocks/shadowsocks-rust`?

madeye: shadowsocks/shadowsocks-rust#211 is not merged yet.

After that, we should change the submodule to upstream.

Mygod: As I mentioned I think that PR doesn't seem too important for this PR right now. :rofl:

madeye:
> This is related to multiple upstreams but we should implement UDP fallback support?

A future plan maybe.

> We should remove hosts setting or add hosts support?

Ditto. For now, I think we can remove this feature. 

> We should remove TFO setting or TFO support?

I suggest remove TFO as well, which is kind of broken in real usage.

> Build speed is unreasonably slow. (use debug profile for debugging and CI builds? enable one target when building debug build? if only the gradle plugin is still being maintained)

Yes, I think we should do that. Previously, I tried to use rust debug target in debug build, but failed.

It's not a bad idea to do some cleanup now. IMO, many features like HOSTS and TFO are not quite useful.

Mygod:
I agree except for UDP fallback because this is the only way to get UDP support for plugins. (UDP is useful for VoIP, gaming, etc and plugin is useful for apparent reasons). Maybe it's best to resolve this before merging it?

madeye:
Okay, let's assume no multi-upstream for now. So, the UDP fallback logic should be the same as before.

Mygod:
Addressed UDP fallback via a252f19 for now (not tested yet) but the real fix should probably be implemented via some kind of multi-upstream, as now we are putting DNS relay inside sslocal.

The rest is probably to address build speed and binary size... (22 minutes of CI build is really slow)

madeye:
Now the CI duration dropped to 10m 8s. https://app.circleci.com/pipelines/github/shadowsocks/shadowsocks-android/8/workflows/8ad4e812-60c7-4aa5-8423-e2d944ac0b20/jobs/526

Mygod:
Oh should probably update license too due to gitmodules changes...

madeye:
UDP relay should work now.

madeye:
Tested locally, fallback UDP relay works now.

madeye:
I think this PR is ready for merge.

Thanks for all the contributions!

tiancaicx:
去除useragent设备标识是如何实现的 这个对突破公司对手机的限制非常有用 客户端希望大神能出win平台的

PR #2455, opened 2020-02-23
Add AlertDialog theme

onlymash:
Fix click effect of dialog buttons under night theme

### DIFF
--- a/plugin/src/main/res/values/styles.xml
+++ b/plugin/src/main/res/values/styles.xml
@@ -8,6 +8,7 @@
         <item name="colorPrimary">@color/color_primary</item>
         <item name="colorPrimaryDark">@color/color_primary_dark</item>
         <item name="windowActionModeOverlay">true</item>
+        <item name="alertDialogTheme">@style/Theme.MaterialComponents.DayNight.Dialog.Alert</item>
     </style>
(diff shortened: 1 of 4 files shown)


Mygod:
Thanks! I have not tested it but shouldn't this be fixed upstream, i.e. at [material-components-android](https://github.com/material-components/material-components-android)?

Mygod:
Tested this patch and it seems like it actually screws up the color.

onlymash:
you mean background color?, customization is easy

Mygod:
I don't see how your patch fixes click effect either. I looked a bit and material-components-android uses `ThemeOverlay` so that you don't need to do this kind of customization. Do you want to update your patch?

plugin/src/main/res/values/styles.xml
```diff
@@ -24,4 +24,7 @@
         <item name="android:windowCloseOnTouchOutside">false</item>
         <item name="windowNoTitle">true</item>
     </style>
+    <style name="AlertDialogTheme" parent="Theme.MaterialComponents.DayNight.Dialog.Alert">
```
Mygod: `ThemeOverlay` should probably be used instead to not screw up the color scheme.

Mygod:
Closing due to inactivity.

PR #2490, opened 2020-04-11
Migrate to Kotlin DSL

onlymash:


Mygod:
I've been meaning to implement this so this is great, but could you make build pass first?

onlymash:
```
error computing cache key: template: 
cacheKey:1:8: executing "cacheKey" at <checksum "build.gradle">: 
error calling checksum: open /home/circleci/code/build.gradle: 
no such file or directory
```

This is not my fault

onlymash:
Update the circleci config
Screenshot_20200412-082116

Mygod:
Thanks!

plugin/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+}
+
+dependencies {
+    val junitVersion = rootExt.get("junit_version") as String
+    val androidTestVersion = rootExt.get("android_test_version") as String
+    val androidEspressoVersion = rootExt.get("android_espresso_version") as String
+
+    api(kotlin("stdlib-jdk8", KotlinCompilerVersion.VERSION))
```
Mygod, line 54: Shouldn't this use `kotlinVersion`?

onlymash: KotlinCompilerVersion.VERSION same as kotlinVersion

Mygod:
Could you merge master and resolve the conflicts?

Also please revert 0ef147b8583a0c3a5acac946f9e0839650a118d0, if any lint check fails I will take care of that for you.

build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+tasks.register("clean", Delete::class) {
+    delete(rootProject.buildDir)
+}
+
+tasks.withType<com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask> {
+    checkForGradleUpdate = true
+    outputFormatter = "json"
+    outputDir = "build/dependencyUpdates"
+    reportfileName = "report"
+}
```
Mygod, line 55: What is this for?

onlymash: https://github.com/ben-manes/gradle-versions-plugin

Mygod: All of these seem optional. I think the default behavior is reasonable.

plugin/build.gradle.kts
Mygod: I don't think so. It seems to be decided by gradle but I will handle it.

PR #2563, opened 2020-07-17
Drop support for Android Lollipop

Mygod:
Alternative fix https://github.com/shadowsocks/shadowsocks-rust/pull/278.


Mygod:
Merging until there is a better fix.

PR #2584, opened 2020-09-28
add share dns services over lan for mobile interface

contrun:
The mobile interface does not have a switch for share over lan which
is presented in the tv interface. Currently sharing dns services over
lan only works if bypass lan is set. Sharing socks5 port works
regardless route config. This is currently not documented and thus may
make some  user confused. I am not knowledgable enough to make sharing
dns services work like sharing other socks5 proxy. Any comments
are welcome.


contrun:
Care to clarify why sharing over lan not presenting in the monile version is expected?

Mygod:
Because users will turn them on accidentally. Considering you carry your phone everywhere, it is not a good idea to expose the port. Thanks for your contribution.

PR #2587, opened 2020-10-05
Avoid text truncation in the StatsBar

### LINKED ISSUE #2492: "HTTPS handshake took .." info cannot be wrapped around to 2nd line
**Describe the bug**

**To Reproduce**
1. Open Shadowsocks app
2. Click on bottom of the screen "Connected, tap to check ..."
3. "Success: HTTPS handshake took ..." info cannot display the 2nd line

**Expected behavior**
The info should be displayed in full.

**Screenshots**
20200413_142125 - Copy
20200413_142126 - Copy

**Smartphone (please complete the following information):**
 - Device: Samsung S20
20200413_1421257 - Copy

**Configuration**
# I don't believe configuration is related here.
Put an `x` inside the [ ] that applies.

* [ ] IPv4 server address
* [ ] IPv6 server address
* [ ] Client IPv4 availability
* Encrypt method:
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [ ] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [ ] IPv6 route
* [ ] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS over UDP
* [ ] Auto Connect
* [ ] TCP Fast Open

**Additional context**
Issue should be related to system font size/zoom. If changed the font to minimum, the info can be displayed on a single line. However it should be able to show up properly even on two (multiple) lines.

Transfusion:
Fixes #2492 

It seems as if the original intention in this commit https://github.com/shadowsocks/shadowsocks-android/commit/3b6cd416939b14c66e18c260dff0272fa366f5a5 was to use `android:ellipsize="marquee"`, which I think is better than removing `android:maxLines="1"` because if the system zoom and font is set to be very large, the `ServiceButton` may still possibly overlap the 2nd row if there is 2 rows and the 1st row has ads. (and furthermore, I remember the text being marquee when this project was still written in Scala...)

Before
image
After
image

### DIFF
--- a/mobile/src/main/java/com/github/shadowsocks/widget/StatsBar.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/widget/StatsBar.kt
@@ -72,7 +72,6 @@ class StatsBar @JvmOverloads constructor(context: Context, attrs: AttributeSet? 
     }
 
     private fun setStatus(text: CharSequence) {
-
         TextViewCompat.setAutoSizeTextTypeWithDefaults(statusText,
                 TextViewCompat.AUTO_SIZE_TEXT_TYPE_NONE)
         statusText.setTextSize(TypedValue.COMPLEX_UNIT_SP, 16f)
--- a/mobile/src/main/java/com/github/shadowsocks/widget/StatsBar.kt
+++ b/mobile/src/main/java/com/github/shadowsocks/widget/StatsBar.kt
@@ -23,11 +23,13 @@ package com.github.shadowsocks.widget
 import android.content.Context
 import android.text.format.Formatter
 import android.util.AttributeSet
+import android.util.TypedValue
 import android.view.View
 import android.widget.TextView
 import androidx.activity.viewModels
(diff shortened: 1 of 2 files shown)


Mygod:
Thanks but I do not like marquee. I wonder if there is a better way to resolve this.

Transfusion:
> Thanks but I do not like marquee. 
I agree. How about this?

image
image
image

Transfusion:
@Mygod Sorry to ping. How is the review going? (I have force-pushed with another implementation based on `autoSize` ing the `TextView`; no longer using marquee).

Mygod:
I think this is the old behavior before I decided to just truncate the text (it is not important anyway). I think the fact that the button is moving up and down annoys me more. Maybe using marquee is the best thing to do...

Mygod:
Closing due to inactivity. Also tap and hold will show you the full text.

PR #2599, opened 2020-11-08
update gradle

Goooler:


Mygod (changes requested):
There are a lot of changes in this one single PR and I do not see the benefits in most of them (e.g. change in code style). Thanks for the contribution but I do not think we will merge this.

PR #2622, opened 2020-12-11
Update shadowsocks-rust to 1.9.0

madeye:


madeye:
The i386 build seems to have problems with illegal instructions.

I guess it's related to crypto2: https://github.com/shadowsocks/crypto2/blob/master/src/blockcipher/aes/x86.rs#L13

For some old x86 devices without AES instructions, we may need to disable the ASM here.

To simplify the logic, we can totally disable assembly acceleration on armeabi and i386 build.

```
/data/user_de/0/com.github.shadowsocks/no_backup/stat_main -c /data/user/0/com.github.shadowsocks/no_backup/shadowsocks.conf --vpn -U --dns-addr 127.0.0.1:5450 --remote-dns-addr dns.google:53 --acl /data/user_de/0/com.github.shadowsocks/no_backup/bypass-lan.acl
12-11 16:58:39.620 11047 11047 F libc    : Fatal signal 4 (SIGILL), code 2 (ILL_ILLOPN), fault addr 0x5f62c2e7 in tid 11047 (libsslocal.so), pid 11047 (libsslocal.so)
12-11 16:58:39.657 11053 11053 I crash_dump32: obtaining output fd from tombstoned, type: kDebuggerdTombstone
12-11 16:58:39.660  1873  1873 I /system/bin/tombstoned: received crash request for pid 11047
12-11 16:58:39.672 11053 11053 I crash_dump32: performing dump of process 11047 (target tid = 11047)
12-11 16:58:39.672 11053 11053 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
12-11 16:58:39.672 11053 11053 F DEBUG   : Build fingerprint: 'google/sdk_gphone_x86/generic_x86:10/QSR1.190920.001/5891938:user/release-keys'
12-11 16:58:39.672 11053 11053 F DEBUG   : Revision: '0'
12-11 16:58:39.672 11053 11053 F DEBUG   : ABI: 'x86'
12-11 16:58:39.672 11053 11053 F DEBUG   : Timestamp: 2020-12-11 16:58:39+0800
12-11 16:58:39.673 11053 11053 F DEBUG   : pid: 11047, tid: 11047, name: libsslocal.so  >>> /data/app/com.github.shadowsocks-aP4h93XpVwl34bjFzFASdQ==/lib/x86/libsslocal.so <<<
12-11 16:58:39.673 11053 11053 F DEBUG   : uid: 10134
12-11 16:58:39.673 11053 11053 F DEBUG   : signal 4 (SIGILL), code 2 (ILL_ILLOPN), fault addr 0x5f62c2e7 (*pc=0x8328f8c5)
12-11 16:58:39.673 11053 11053 F DEBUG   :     eax ffb366f4  ebx 5f9abe24  ecx ef469068  edx 00000000
12-11 16:58:39.673 11053 11053 F DEBUG   :     edi ffb366f4  esi 0000000d
12-11 16:58:39.673 11053 11053 F DEBUG   :     ebp ffb366f4  esp ffb36630  eip 5f62c2e7
12-11 16:58:39.685 11053 11053 F DEBUG   :
12-11 16:58:39.685 11053 11053 F DEBUG   : backtrace:
12-11 16:58:39.685 11053 11053 F DEBUG   :       #00 pc 000cb2e7  /data/app/com.github.shadowsocks-aP4h93XpVwl34bjFzFASdQ==/lib/x86/libsslocal.so
12-11 16:58:39.685 11053 11053 F DEBUG   :       #01 pc 000898e8  /apex/com.android.runtime/lib/bionic/libc.so (__libc_init+120) (BuildId: 76290498408016ad14f4b98c3ab6c65c)
```

zonyitoo:
I am still working on https://github.com/shadowsocks/shadowsocks-rust/issues/326 , which is the actual v1.9.0.

madeye:
Sure, no hurry.

LuoZijun:
@madeye 我想问题应该在这个文件里：https://github.com/shadowsocks/shadowsocks-rust/blob/master/.cargo/config.toml#L10

如果你为一些陈旧的处理器编译版本的话，那可能需要修改这个 RUSTFLAGS，把 AES 和 SSE2 去掉。
或者直接在环境变量里面使用 `RUSTFLAGS="-C -C target-feature=+aes,+pclmul,+sse,+sse2,+sse3,+ssse3,+sse4.1,+sse4.2,+avx,+avx2" cargo build --release`

madeye:
@LuoZijun We'd better separate x86 and x86_64 and enable aes and avx for x86_64 only.

I'll try overwrite RUSTC flags in the Gradle config.

LuoZijun:
@madeye 

不对，这应该是 Ring 的问题，因为 `shadowsocks-crypto` 在 `X86`、`X86-64`、`ARM32`、`AArch64` 都是使用 Ring 的密码代码（AEAD）。

Ring 的 `build.rs` 脚本可以看到，Ring 在 X86 平台依然适用 AES-NI 指令 以及 `pclmulqdq` 指令。

https://github.com/briansmith/ring/blob/main/build.rs#L55-L60

Mygod (commented):
I think it's better to wait for the features in rust-nightly become stable before merging this. (I haven't been following -- what features were required on nightly?) Otherwise looks good.

madeye:
It looks ring is still in a early stage. I see some similar issues for armv7 build:

https://github.com/briansmith/ring/issues/838

madeye:
If we plan to switch to ring and crypto2 in the next release, we may need to remove the support of armv7 and x86.

Mygod:
I think you might be heavily underestimating how popular 32-bit systems are due to memory constraints.

madeye:
Guess we need a LTS branch from shadowsocks-rust built with openssl and libsodoum...

Or we should find a way to disable AES and AVX for x86 and armv7 target.

zonyitoo:
> Guess we need a LTS branch from shadowsocks-rust built with openssl and libsodoum...
> 
> Or we should find a way to disable AES and AVX for x86 and armv7 target.

Maybe we could simply disable `ring` on x86 and armv7 targets?

madeye:
> Maybe we could simply disable ring on x86 and armv7 targets?

It's a good idea.

madeye:
BTW, is there any way to optimize the memory usage for ACL. I observed very large memory usage from the previous release. The memory usage easily reaches to 200M in several minutes after the startup.

```
 9969 u0_a446      20   0  12G 197M 2.3M S 18.0   2.6   0:25.90 libsslocal.so +
```

I suspected it's related to the regex we're using now. Maybe we should move to wildcard rules to solve this issue.

Mygod:
@madeye Make a separate thread? Has the issue been here since the beginning?

zonyitoo:
> BTW, is there any way to optimize the memory usage for ACL. I observed very large memory usage from the previous release. The memory usage easily reaches to 200M in several minutes after the startup.
> 
> ```
>  9969 u0_a446      20   0  12G 197M 2.3M S 18.0   2.6   0:25.90 libsslocal.so +
> ```
> 
> I suspected it's related to the regex we're using now. Maybe we should move to wildcard rules to solve this issue.

Yes. I have also noticed that on my RPI. Regex rules matching is very CPU consuming, which results in very high latency on ARM devices.

zonyitoo:
Is it possible to provide local DNS resolver with common TCP & UDP sockets? That would make a lot easier to reuse those existed DNS resolvers.

LuoZijun:
@zonyitoo See https://github.com/shadowsocks/dns/blob/main/wire/examples/lookup.rs

zonyitoo:
> @zonyitoo See https://github.com/shadowsocks/dns/blob/main/wire/examples/lookup.rs
> 
> 也许对你有帮助。

The key issue is that Android is currently using an unix socket as local DNS resolver, which is not compatible with common DNS libraries. Usually these libraries contains lots of useful tools, like

- RFC requires resolver to handles `localhost` as a special case
- Caching query results with `ttl`
- DNS over TLS / HTTPS, ... mdns, ... dnssec, and so on.

madeye:
@zonyitoo I suggest we keep the current RPC like local resolver for Android app.

Actually, the local resolver provided by shadowsocks-android is just a wrapper of Android' API: https://developer.android.com/reference/java/net/InetAddress#getAllByName(java.lang.String). We have no plan to implement a real local DNS resolver in Kotin or Java.

We expect all the local resolver logic would be moved to shadowsocks-rust one day. It may not an easy task, as we need to find a way to call this API from shadowsocks-rust: https://developer.android.com/ndk/reference/group/networking#group___networking_1ga0ae9e15612e6411855e295476a98ceee

madeye:
I added a trade-off via 122ed269.

Now shadowsocks-android's local DNS wrapper listening on local UDP port instead of the UNIX domain socket. It works with `local-dns` well. Not sure if it also works `trust-dns`.

zonyitoo:
That's great. Test steps:

1. Delete [these line of codes](https://github.com/shadowsocks/shadowsocks-rust/blob/a7e0478dd4f7cc903fd9aed6fdb59067e8ee4a9a/crates/shadowsocks-service/src/local/mod.rs#L70-L81), which will uses `trust-dns-resolver` as default DNS resolver (just for resolving servers' addresses)
2. Set `dns` key in configuration file:
    ```json
    {
        "dns": "127.0.0.1:PORT",
    }
    ```
    This will set the name server configuration for `trust-dns-resolver`, it will connects to this address with both TCP and UDP protocol.

madeye:
@zonyitoo What's the release plan for shadowsocks-rust 1.9.0? I guess including tokio 1.0 in 1.9.0 is too risky?

zonyitoo:
> @zonyitoo What's the release plan for shadowsocks-rust 1.9.0? I guess including tokio 1.0 in 1.9.0 is too risky?

It should include tokio v1.0. Because it is almost equivalent to v0.3. All libraries out there held release and waiting for v1.0.

madeye:
> This will set the name server configuration for trust-dns-resolver, it will connects to this address with both TCP and UDP protocol.

It looks not working. All local DNS queries are dropped.

> It should include tokio v1.0. Because it is almost equivalent to v0.3. All libraries out there held and waited for v1.0.

Okay.

zonyitoo:
> It looks not working. All local DNS queries are dropped.

```
"--local-dns-addr", "127.0.0.1:${DataStore.portLocalDns + 1}"
```

This command line option is also required.

madeye:
Okay, it works now.

zonyitoo:
Great. Did you delete those code before test?

madeye:
> Did you delete those code before test?

Yes

zonyitoo:
BTW. On the latest master branch, I have made some modifications about dns-relay:

For remote DNS queries, it sends TCP and UDP queries simutaneously. But TCP queries sends ramdonly 500ms ~ 1.5s after UDP queries. https://github.com/shadowsocks/shadowsocks-rust/blob/9bc90cc59bc9a1fe61cdab47648c02b055878e5a/crates/shadowsocks-service/src/local/dns/server.rs#L633-L670

zonyitoo:
> We expect all the local resolver logic would be moved to shadowsocks-rust one day. It's not an easy task, as we need to find a way to call this API from shadowsocks-rust: https://developer.android.com/ndk/reference/group/networking#group___networking_1ga0ae9e15612e6411855e295476a98ceee

https://github.com/bbqsrc/cargo-ndk   :P

madeye:
@zonyitoo BTW, the local resolver is UDP only. So we need to make the UDP local resolver always enabled even in TCP only mode.

madeye:
> @zonyitoo BTW, the local resolver is UDP only. So we need to make the UDP local resolver always enabled even in TCP only mode.

May we can control the local and remote DNS protocols? For example, `--local-dns-addr udp://127.0.0.1:5353`

madeye:
Moved to DNS changes to https://github.com/shadowsocks/shadowsocks-android/pull/2635

zonyitoo:
> > @zonyitoo BTW, the local resolver is UDP only. So we need to make the UDP local resolver always enabled even in TCP only mode.
> 
> Maybe we can control the local and remote DNS protocols? For example, `--local-dns-addr udp://127.0.0.1:5353`

That's a good idea. But for general cases, a DNS server must provide both UDP and TCP services. trust-dns won't create the TCP stream if UDP queries are alway succeeded.

IceCodeNew:
@madeye 这个 PR 在依赖 tokio 的依赖项还没有兼容 tokio 1.0 的时候就合并了，我在想是不是应该再更新一次 git submodule `shadowsocks-rust`，重新触发一次编译？

Mygod:
> @zonyitoo I suggest we keep the current RPC like local resolver for Android app.
> 
> Actually, the local resolver provided by shadowsocks-android is just a wrapper of Android's API: https://developer.android.com/reference/java/net/InetAddress#getAllByName(java.lang.String). We have no plan to implement a real local DNS resolver in Kotlin or Java.
> 
> We expect all the local resolver logic would be moved to shadowsocks-rust one day. It's not an easy task, as we need to find a way to call this API from shadowsocks-rust: https://developer.android.com/ndk/reference/group/networking#group___networking_1ga0ae9e15612e6411855e295476a98ceee

Incorrect. On latest Android versions, we use per-Network resolving APIs. Will open a separate issue about UDS later.

PR #2635, opened 2020-12-25
Switch to local UDP DNS resolver

madeye:


madeye:
sslocal only sends local UDP queries if UDP mode is enabled:

https://github.com/shadowsocks/shadowsocks-rust/blob/303f96c61f520f447589467339a160ad118011d6/crates/shadowsocks-service/src/local/dns/server.rs#L685-L696

However, shadowsocks-android only provided a UDP local resolver.

@zonyitoo

zonyitoo:
Hmm.. That means I shouldn't check `mode` for local relay?

madeye:
I think the TCP/UDP mode defined here is only for shadowsocks upstream, right?

If so, we should remove this condition for the local resolver.

zonyitoo:
Agree.

.circleci/config.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     environment:
       GRADLE_OPTS: -Dorg.gradle.workers.max=1 -Dorg.gradle.daemon=false -Dkotlin.compiler.execution.strategy="in-process"
     steps:
-      - run: rustup toolchain install nightly-2020-12-20
-      - run: rustup override set nightly-2020-12-20
+      - run: rustup toolchain install nightly-2020-12-17
+      - run: rustup override set nightly-2020-12-17
```
zonyitoo, line 11: Actually you don't need to do this. `rustup` will automatically reads the configuration from `rust-toolchain`

madeye: Okay, let me remove this.

madeye: ···
error[E0463]: can't find crate for `core`
  |
  = note: the `armv7-linux-androideabi` target may not be installed
···

If I remove this, I will get the above error.

madeye: https://app.circleci.com/pipelines/github/shadowsocks/shadowsocks-android/193/workflows/5532d5b6-0863-46f0-b134-51e0e82031c7/jobs/704

madeye: it looks rustup cannot install targets automatically. Any suggestion?

zonyitoo:
```
info: syncing channel updates for 'nightly-2020-12-17-x86_64-unknown-linux-gnu'
info: latest update on 2020-12-17, rust version 1.50.0-nightly (b32e6e6ac 2020-12-16)
```

Hmm..

zonyitoo: The answer is yes. I took a look at [`rust-embedded/cross`](https://github.com/rust-embedded/cross/blob/4f213ba07b214d36225be63e5b72a6346937a699/src/main.rs#L226-L241) and [`actions-rs/toolchain`](https://github.com/actions-rs/toolchain/blob/88dc2356392166efad76775c878094f4e83ff746/src/main.ts#L9-L13), they reads from the `rust-toolchain` file and pass it manually to `rustup toolchain add ...`

zonyitoo: Consider using `rust-embedded/cross`?

madeye: Fixed via https://github.com/shadowsocks/shadowsocks-android/pull/2635/commits/45733069240c41a2b6ac4aa2369ee2996de80139.

PR #2672, opened 2021-02-05
Update dependencies

Goooler:
- Replace `https://plugins.gradle.org/m2/` with `gradlePluginPortal()`, same address.
- Update dependencies include kotlin 1.4.30
- Replace `get` call with indexing operator


Goooler:
- Replace 'get' call with indexing operator

Mygod (approved):

PR #2693, opened 2021-03-28
Add packagingOptions exclude rules

Goooler:
Add rules to exclude types of file when packaging, result like this
image


Mygod:
I am not sure if those files could be excluded. Source?

Goooler:
> I am not sure if those files could be excluded. Source?

No problem, you can have a functional test after that.

Mygod:
Things you are excluding take up less than 1% (70KB) and prevents things like Google Play checks from functioning.

Issue #2678, opened 2021-02-15
A public Maven repo for plugin lib version 2.0.0


tuubrooks:
**Is your feature request related to a problem? Please describe.**
I've been trying to implement and build a plugin using plugin lib version 2.0.0, but it took some time to realize that there isn't a public v2.0.0 plugin repo on https://jcenter.bintray.com.

**Describe the solution you'd like**
Appreciate it very much if plugin lib v2.0.0 could be published to jcenter.bintray.com. 
It will save some time during implementation of a v2 plugin.

**Describe alternatives you've considered**
Of course, a locally self-built lib is always a backup plan, but in a time-consuming manner

PR #2706, opened 2021-04-14
Plugin 2.0.1

Mygod:
Needs more testing. Main change is using new fragment result API from androidx.fragment 1.3.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -68,7 +68,6 @@ dependencies {
     val workVersion = "2.5.0"
 
     api(project(":plugin"))
-    api("androidx.appcompat:appcompat:1.2.0")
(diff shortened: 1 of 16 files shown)


madeye (approved):

Mygod:
We should maybe fix #2678 too. Also [jcenter is being a *responsible* platform and going bye bye](https://developer.android.com/studio/build/jcenter-migration).

madeye:
#2678 should be fixed now. 2.0.0 is uploaded months ago.

Mygod:
Sounds good. This PR is good to go.

mrkoikoi (commented):
https://basic.bravedns.com/1:AIBAAQ==

PR #2731, opened 2021-05-08
Resolve server name dynamically always

### LINKED ISSUE #2722: DDNS support with plugins
Now that we have DDNS from #2301, we should support it when it is used along with a plugin as well. One thing to do is to start a second plugin when IP is changed, and release the first one when all existing connections have been closed.

CC @zonyitoo @madeye

Mygod:
Fixes #2722. This also allows turning on service without having an Internet connection, even when the server is a domain name. Neat!


madeye (approved):

Mygod:
@madeye Could do a new release of the app along with the plugin. :eyes:

PR #2887, opened 2022-05-20
upgrade to SIP022 ciphers

dev4u:
(https://github.com/shadowsocks/shadowsocks-android/discussions/2865)
Maybe really need to release a new version, so more people can test the new ciphering.
The new ciphering may cause the cell phone's battery to be less durable.


Mygod:
@madeye What's your thoughts on this 2022 shenanigans? I haven't been bothered to read it yet.

core/build.gradle.kts
```diff
@@ -39,12 +39,16 @@ cargo {
...
-            "aead-cipher-extra",
-            "logging",
-            "local-flow-stat",
-            "local-dns"))
+        "stream-cipher",
+        "aead-cipher-extra",
+        "logging",
+        "local-flow-stat",
+        "local-dns",
+        "security-iv-printable-prefix",
```
zonyitoo, line 47: This feature should be removed because it is not confirmed yet.

core/build.gradle.kts
```diff
@@ -39,12 +39,16 @@ cargo {
...
+        "aead-cipher-2022",
```
zonyitoo, line 48: Since you are using Rust nightly, consider adding `armv8` and `neon` for producing optimized code for ciphers.

zonyitoo:
Apart from SIP022 (AEAD-2022), there are some improvements in shadowsocks-rust:

1. shadowsocks-crypto's ciphers have been switched to RustCrypto community, which is in more active development. But RustCrypto's implementations are slightly slower than crypto2.
2. An experimental `tun` interface support. It still need performance improvement but working well in most cases.
3. Supporting `|` and `||` in ACL rules for better performance.

AEAD-2022 is ready for test, what do you think about this new protocol? @Mygod

Mygod:
I am unconvinced that a backwards-incompatible upgrade is necessary.

dev4u:
2022协议能走多远？这没人能预测。向后兼容，这更无从说起。
一套协议肯定不是一蹴而就，要经过不断修改、验证、再修改……主要方向对了，那就开干吧。
shadowsocks的协议，也不是一开始就是现在这样，中间也有翻阴沟的时候，例如ota就直接被舍弃。
SS得以各位大佬的奉献，让SS得以持续向前发展。
2022协议，也是为了修复现有协议的缺陷才设计出来的。经过我个人试用后，仍存在某些缺陷。就算是这样，个人觉得还是应该让更多的人参与进来。只有扩大测试人群，得到足够的反馈，才能验证新协议是否有效。

zonyitoo:
> I am unconvinced that a backwards-incompatible upgrade is necessary.

Some discussions in shadowsocks-org have shown that there may be no compatible way to resolve the security flaws. Proposal in SIP022 might be the minimal modification against the SIP004 AEAD protocol.

In the UDP protocol, the main goal is to reduce cryptography overhead, so SIP002 introduced a simple session based protocol. It performs well in stress test.

It is not perfect, maybe. So I will still keep shadowsocks-rust v1.15.0 in alpha release. shadowsocks-android may make an alpha release for letting more users to join for testing.

madeye:
@Mygod no matter what I think, they decided to add this cipher... Maybe we just follow it and do some experimental releases, in case anyone wants to give it  a try.

madeye:
Attached an experimental release here:

https://www.dropbox.com/s/unp2vy2zj5vi8yo/shadowsocks-aead-2022.zip

dev4u:
有点难为你了，大兄弟。

dev4u:
如果有小伙伴构建这个PR的，请将ss rust依赖版本更新至：https://github.com/shadowsocks/shadowsocks-rust/commit/c9f7a7b8aed8c5209d241cfe2982015dd899d423
ref https://github.com/shadowsocks/shadowsocks-rust/issues/855

core/src/main/java/com/github/shadowsocks/bg/TrafficMonitor.kt
```diff
@@ -69,8 +69,8 @@ class TrafficMonitor(statFile: File) {
         if (delta != 0L) {
             if (dirty) {
                 out = current.copy().apply {
-                    txRate = (txTotal - out.txTotal) * 1000 / delta
-                    rxRate = (rxTotal - out.rxTotal) * 1000 / delta
+                    txRate = (txTotal - out.txTotal) * 1024 / delta
+                    rxRate = (rxTotal - out.rxTotal) * 1024 / delta
```
Mygod, line 73: This is s to ms.

core/src/main/res/values/arrays.xml
```diff
@@ -2,6 +2,16 @@
 <resources>
     <string-array name="enc_method_entry" translatable="false">
         <item>NONE</item>
+        <item>CHACHA20-IETF</item>
+        <item>AES-128-GCM</item>
+        <item>AES_128_GCM_SIV</item>
+        <item>AES-256-GCM</item>
+        <item>AES_256_GCM_SIV</item>
```
Mygod, line 9: Why inconsistent case?

core/src/main/java/com/github/shadowsocks/bg/TrafficMonitor.kt
dev4u: 🤣是我贻笑大方了。

Issue #2930, opened 2022-08-30
Exit code:9009


newfuture33:
When build code, faced this issue. I have found this file in position. File exist, but get 9009 error.
![image](https://user-images.githubusercontent.com/34529419/187488312-b052cc2a-af62-463f-bd58-479ad3237b67.png)

xiedeacc:
same problem
os:win10
android studio: 2021.2.1 patch2
rustc:1.61
android sdk build-tools:33

xiedeacc:
it seems python problem???
compiled success by change rustc to nightly on ubuntu, but still failed on windows

jeytin:
I  maybe  solve it，it seems python3 command  not found .  

```kotlin 

cargo {
    module = "src/main/rust/shadowsocks-rust"
    libname = "sslocal"
    //targets = listOf("arm", "arm64", "x86", "x86_64")
    targets = listOf("arm64")
    profile = findProperty("CARGO_PROFILE")?.toString() ?: currentFlavor
    extraCargoBuildArguments = listOf("--bin", libname!!)
    featureSpec.noDefaultBut(arrayOf(
        "stream-cipher",
        "aead-cipher-extra",
        "logging",
        "local-flow-stat",
        "local-dns",
        "armv8",
        "neon",
        "aead-cipher-2022",
    ))
    exec = { spec, toolchain ->
        spec.environment("RUST_ANDROID_GRADLE_PYTHON_COMMAND", "python")
        spec.environment("RUST_ANDROID_GRADLE_LINKER_WRAPPER_PY", "$projectDir/$module/../linker-wrapper.py")
        spec.environment("RUST_ANDROID_GRADLE_TARGET", "target/${toolchain.target}/$profile/lib$libname.so")
    }
}

````

 “python3”  modify to  “python”，no  this error,  you can try.

PR #2935, opened 2022-09-17
Fix the compile error #2930

dev4u:


core/build.gradle.kts
```diff
@@ -51,9 +51,17 @@ cargo {
         "aead-cipher-2022",
     ))
     exec = { spec, toolchain ->
-        spec.environment("RUST_ANDROID_GRADLE_PYTHON_COMMAND", "python3")
-        spec.environment("RUST_ANDROID_GRADLE_LINKER_WRAPPER_PY", "$projectDir/$module/../linker-wrapper.py")
-        spec.environment("RUST_ANDROID_GRADLE_TARGET", "target/${toolchain.target}/$profile/lib$libname.so")
+        run {
+            val process: Process = Runtime.getRuntime().exec("which python3 >/dev/null 2>&1 && echo python3 installed")
```
madeye, line 55: I don't think 'which' can work on Windows?

dev4u: 噢，我改一下。

dev4u:
@madeye 已经修改，请检视。

madeye:
I didn't really get the idea of this fix. You are assuming python or python3 should be installed?

Maybe output an error message is more friendly.

dev4u:
好的，我明白你的意思。
我回去再改一下。

dev4u:
@madeye how about this latest commit?

madeye:
merged, thanks.

Issue #2606, opened 2020-11-27
No Internet Access: Unable to resolve host "cp.cloudflare.com"


asycns:
**Describe the bug**
When I test the connection, I get "No Internet access. Unable to resolve host 'cp.cloudflare.com', No address associated with hostname".
It errors most of the time, but my laptop and my iPad works normally. 
DNS server works properly. The server works well. Settings are all correct. No block. 
Several other WiFi networks are tried, but it stills. 
Network settings are cleared. 

**To Reproduce**
1. Connect to Shadowsocks server
2. Test the connection
3. Failed and reported error

**Expected behavior**
Connect to Shadowsocks server 

**Screenshots**
![image](https://user-images.githubusercontent.com/46381867/100469527-55493c00-3111-11eb-854d-ef60026ee789.png)

**Smartphone (please complete the following information):**
 - Device: iQOO z1x
 - Version: Android 10.0; Shadowsocks 5.1.4 & 5.0.6 from play store
 - Last version that did not exhibit the issue: No

**Configuration**
Put an `x` inside the [ ] that applies.

* [x] IPv4 server address
* [ ] IPv6 server address
* [x] Client IPv4 availability
* [ ] Client IPv6 availability
* Encrypt method: chacha20-ietf-poly1305
* Route
  * [ ] All
  * [ ] Bypass LAN
  * [ ] Bypass China
  * [x] Bypass LAN & China
  * [ ] GFW List
  * [ ] China List
  * [ ] Custom rules
* [x] IPv6 route
* [ ] Apps VPN mode
  * [ ] Bypass mode
* Remote DNS: 8.8.8.8
* [ ] DNS over UDP
* Plugin configuration (if applicable):
* [ ] Auto Connect
* [ ] TCP Fast Open
* If you're not using VPN mode, please supply more details here:

**Additional context**
It reproduces on my two vivo phone but fails to reproduce on my xiaomi phone. My xiaomi 6 (Android 9.0) works pretty well.

asycns:
In addition, none of the route works in my vivo phones.

Celthi:
I met the same error.

lavi-xu:
A simple way to reproduce:
Run server on Windows 10 : gost.exe -L=ss://aes-256-cfb:test1234@:8388
Download binary here: https://github.com/ginuerzh/gost/releases/tag/v2.11.1

Run android and configure the same, do test. You will see "Unable to resolve host "cp.cloudflare.com".

[...]

PR #2950, opened 2022-11-07
Changing test url from cp.cloudflare.com

akajaber:
Hi,
Due to the fact that a large part of the requests sent to cloudflare.com are blocked in some countries such as Iran, in many cases it is not possible to test the connection for users and thus it is not possible to connect to the server.

One of the most suitable URLs, which practically cannot be limited due to the use of a huge amount of users, is :

`http://detectportal.firefox.com/success.txt.`

Therefore, most of the users' problems will be solved through

### DIFF
--- a/core/src/main/java/com/github/shadowsocks/net/HttpsTest.kt
+++ b/core/src/main/java/com/github/shadowsocks/net/HttpsTest.kt
@@ -82,7 +82,7 @@ class HttpsTest : ViewModel() {
     fun testConnection() {
         cancelTest()
         status.value = Status.Testing
-        val url = URL("https://cp.cloudflare.com")
+        val url = URL("http://detectportal.firefox.com/success.txt")
         val conn = (if (DataStore.serviceMode != Key.modeVpn) {
             url.openConnection(Proxy(Proxy.Type.SOCKS, DataStore.proxyAddress))
         } else url.openConnection()) as HttpURLConnection


akajaber:
> It is crucial that we use `https` to reflect "real Internet access". Firefox's server is instead used to detect captive portals which is a different purpose.

Thanks for your reply, You made the right point

The main issue here is that countries that have internet censorship systems block many CDN services such as Cloudflare and requests related to their IPs and domains. which has been discussed in various issues regarding this problem. issue #2606 & issue #2705 

Just by changing cp.cloudflare.com to any other similar service, it can get exactly the desired result in the application, and in this way, the problems of many users will be solved.

Even in this case, instead of **Firefox Detect Portal**, we can use another item such as:
`https://clients3.google.com/generate_204`
used.

The importance note about this, is that the TestUrl changes to something other than cloudflare.

akajaber (commented):
Alternatives for testConnection url : 
```
https://detectportal.firefox.com/success.txt // SSL support
https://clients3.google.com/generate_204
```

ElderArcade:
Alternatives for testConnection url:
http://gstatic.com/generate_204
http://clients3.google.com/generate_204
http://cp.cloudflare.com
http://detectportal.firefox.com/success.txt


https://gstatic.com/generate_204
https://clients3.google.com/generate_204
https://cp.cloudflare.com
https://detectportal.firefox.com/success.txt

PR #3116, opened 2023-06-19
Add "Hide On Recents Screen" Option

guxin0123:
Add "Hide On Recents Screen" Option

Can prevent being mistakenly terminated


Mygod:
Thanks for contributing! However, manipulating recents doesn't seem like a good idea and seems too niche. Therefore, I do not recommend merging.

PR #3231, opened 2026-02-12
Remove ABI splits in favor of AAB delivery

madeye:
## Summary
- Remove `splits.abi` block from `Helpers.kt` — ABI splits are incompatible with AAB (`bundleRelease` fails with "Multiple shrunk-resources files found"), and AAB already handles per-device ABI delivery natively
- Update `release.sh` to copy AAB outputs instead of per-ABI APKs
- Fix APK paths in E2E test and CI scripts to use universal APK name

## Test plan
- [x] `./gradlew assembleDebug` builds successfully
- [x] `./gradlew :mobile:bundleRelease` builds successfully
- [x] E2E tests pass (4/4: tun0, DNS, TCP x2)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -70,7 +70,7 @@ jobs:
         env:
           SKIP_EMULATOR_BOOT: "true"
           ADB: adb
-          APK: ${{ github.workspace }}/mobile/build/outputs/apk/debug/mobile-x86_64-debug.apk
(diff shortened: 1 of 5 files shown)


Mygod:
🤔 Since we are not releasing on Google Play, it might be worth keeping either a single universal apk, or just split only and ask non-tech-savvy users to use Obtainium to install this app.

PR #3244, opened 2026-05-08
Fix profile JSON import broken by trailing newline

trulyliu:
The gson→org.json migration in df4d65cb wrapped the root-value read in `while (JSONTokener.more()) process(nextValue())`, but JSONTokener.more() reports true on trailing whitespace. Every JSON file ending with a newline (i.e. essentially every file) hits a second nextValue() call that runs off the end and throws JSONException, before finalize() persists the parsed profile.

Restore the pre-migration single-value semantic.


trulyliu:
This PR fixes a bug in the latest version that prevents the import of JSON configuration files on TV.

Mygod (changes requested):
This changes the functionality.

